PR #2, opened 2018-03-28
Export types

billneff79:


types.ts
```diff
@@ -1,8 +1,6 @@
 export * from './src/schema/generated-schema-types';
 export { ActionType, ActionResultType } from './src/batch-client/types';
 
-export const ReadingPaneSashHorizontalDefault = 50;
```
billneff79: This is a UI constant, not an API constant.

PR #3, opened 2018-03-28
PREAPPS-302 add change password and logout functionality

billneff79:


billneff79:
This is really @pl12133 's PR.  I approved and am merging.

PR #5, opened 2018-04-11
PREAPPS-456: remove some line of code which was not passing errors to component as response

lavegupta:


src/apollo/zimbra-error-link.ts
```diff
@@ -1,22 +1,12 @@
...
-		) {
-			graphQLErrors.map(({ message, path }) =>
-				console.log(`[Unsupported API]: ${message}, Path: ${path}`)
-			);
-			if (response) {
-				response.errors = undefined;
-			}
-		}
+const ZimbraErrorLink: ApolloLink = onError(({ graphQLErrors }) => {
+	if (graphQLErrors) {
```
billneff79: The whole point of this ApolloLink was to not respond with errors for unsupported APIs.  Since we don't know which version of Zimbra we are going to be hitting, I don't think we know that up front like the `PARTIAL_SUPPORT` api suggests.  I think it's better, as you are doing, to pass the errors through and let the client worry about handling it.  

With that said, this error link doesn't even need to exist anymore.  You can just delete the file and remove its existence in the global exports and remove its usage here: https://github.com/Zimbra/zm-x-web/blob/8837b2f39dad40e5bad65e7218410e29d91d215c/src/components/app/index.js#L140

PR #6, opened 2018-04-11
Adjust settings to `uglifyjs` mangler to allow CJS and UMD bundles to build correctly.

pl12133:
Keep the `_createClass` name to avoid it being removed by dead code elimination.
Keep function names because `graphql` relies on them.


billneff79:
Do you need a new version released for this now, or do we want to wait for the next necessary version push?

pl12133:
I'm not aware of anyone being effected by this broken build problem so it doesn't need to be in an immediate release.

pl12133:
It also seems like using the CJS or UMD bundles will trigger https://github.com/graphql/graphql-js/issues/594 in the `zm-x-web` application, presumably because there are two versions of `graphql`: the version bundled into `@zimbra/api-client` conflicting with the version existing in the main `zm-x-web` app. I could be wrong because I haven't taken much time to investigate.

Since building from the ESM bundle can deduplicate the imports, it doesn't show the same problem; and that makes it a low priority to fix so I will not be looking into it further until someone presents it as a real problem.

PR #7, opened 2018-04-12
PREAPPS-264 Add new sort by enum values to schema

billneff79:
and Add auto-typescript generation from graphql schema

I used the `graphql-code-generator` package to generate the typescript files from the interpolated schema json because that's something that apollo-codegen just won't do oddly enough (see: https://github.com/apollographql/apollo-codegen/issues/351 )

The generated file looks quite a bit different, but that's just because a different template is being used to generate it, and any types that came from the `zm-x-web` repo are no longer there.  I manually verified that the types needed by the zm-api-js-client repo exists, and that this works/build zm-x-web still.

This should ensure that anytime we build/publish this repo, the exported types are automatically updated from the schema. 🎉


src/schema/schema.ts
```diff
@@ -64,12 +62,10 @@ export function createZimbraSchema(
 					client.getContactFrequency(variables as GetContactFrequencyOptions),
 				getConversation: (_, variables) =>
 					client.getConversation(variables as GetConversationOptions),
-				getFolder: (_: any, variables: getFolderQueryVariables) =>
```
billneff79: This was unnecessarily using enums from graphql files in the zm-x-web repo in the old version of the generated types file.

PR #9, opened 2018-04-17
PREAPPS-548 Remove unused and shouldn't be used calendars and folders queries

billneff79:
This should not be published until https://github.com/Zimbra/zm-x-web/pull/35 is merged, which stops using these in favor of the standard `getFolder` query

### DIFF
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -195,10 +195,6 @@ Folder.addMapping({
 });
 export { Folder };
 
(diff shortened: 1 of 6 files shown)


src/normalize/entities.ts
```diff
@@ -195,10 +195,6 @@ Folder.addMapping({
 });
 export { Folder };
 
-export const GetFolderResponse = new Entity({
```
billneff79: This normalization is already done by the `Folder` entity

PR #14, opened 2018-05-15
Preapps 503 Edit appointments

silentsakky:
- Open appointment edit view on double click
- add context menu for appointments with edit option
- Add edit button on appointment details tooltip
- create graphql decorators for create/update/delete appointment
- modify graphql schema to add alarm data as part of InviteComponent
- add functionality to create exception in recurring series
- add support for editing of appointments
- add support for showing exceptions in calendar


src/normalize/entities.ts
```diff
@@ -54,8 +50,7 @@ const RecurrenceInfo = new Entity({
 });
 
 const CalendarItemAlarm = new Entity({
-	trigger: CalendarItemAlarmTrigger,
-	at: ['attendees', CalendarItemAlarmAttendees]
```
pl12133: Any reason to remove this? It seems like a valid property on an `<alarm>`.

billneff79: - [ ] let's put this back.  It's in the graphql schema for the output

src/schema/schema.graphql
```diff
@@ -201,8 +201,27 @@ type Instance {
 	start: Float
 	dueDate: Float
 	tzoDue: Int
-	isRecurring: Boolean
```
billneff79: Either this needs to stay in, or we should stop denormalizing it in the entity definition above

src/schema/schema.graphql
```diff
@@ -864,6 +886,12 @@ input CalendarItemDateTimeInput {
 	date: String!
 }
 
+# @FIXME
```
billneff79: - [ ] Remove this comment

src/schema/schema.graphql
silentsakky: Removed from normalisation logic

src/schema/generated-schema-types.ts
```diff
@@ -480,8 +498,26 @@ export interface Instance {
 	start?: number | null;
 	dueDate?: number | null;
 	tzoDue?: number | null;
-	isRecurring?: boolean | null;
 	utcRecurrenceId?: string | null;
+	alarm?: boolean | null;
```
billneff79: Do we really need all of these ` | null`s?  What is that accomplishing for us?

billneff79: nevermind - this is the generated file :)

PR #15, opened 2018-05-22
Preapps 632 Saving filters should not overwrite existing filters

silentsakky:
- add support for updating filters through graphql
- add all schema for filters which are supported by legacy client

### DIFF
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -263,3 +263,69 @@ export const SearchResponse = new Entity({
 	c: ['conversations', Conversation],
 	cn: ['contacts', Contact]
 });
+
+const RedirectAction = new Entity({
+	a: 'address'
+});
+
+const NotifyAction = new Entity({
+	a: 'address',
+	su: 'subject'
+});
+
+const FilterAction = new Entity({
+	actionKeep: 'keep',
+	actionDiscard: 'discard',
+	actionFileInto: 'fileInto',
+	actionFlag: 'flag',
+	actionTag: 'tag',
+	actionRedirect: ['redirect', RedirectAction],
+	actionReply: 'reply',
+	actionNotify: ['notify', NotifyAction],
+	actionStop: 'stop'
+});
+
+const DateCondition = new Entity({
+	d: 'date'
+});
+
+const ImportanceCondition = new Entity({
+	imp: 'importance'
+});
+
+const SizeCondition = new Entity({
+	s: 'size'
+});
+
+const FilterCondition = new Entity({
+	condition: 'allOrAny',
+
+	addressBookTest: 'addressBook',
+	addressTest: 'address',
+	attachmentTest: 'attachment',
+	bodyTest: 'body',
+	bulkTest: 'bulk',
+	contactRankingTest: 'contactRanking',
+	conversationTest: 'conversation',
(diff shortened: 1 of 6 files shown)


src/normalize/entities.ts
```diff
@@ -263,3 +263,69 @@ export const SearchResponse = new Entity({
...
+});
+
+const SizeCondition = new Entity({
+	s: 'size'
+});
+
+const FilterCondition = new Entity({
+	condition: 'allOrAny',
+
+	addressBookTest: 'addressBook',
```
billneff79: What was the importance of removing "Test" from the name of each filter condition?

src/normalize/entities.ts
```diff
@@ -263,3 +263,69 @@ export const SearchResponse = new Entity({
...
+	mimeHeaderTest: 'mimeHeader',
+	sizeTest: ['size', SizeCondition],
+	twitterTest: 'twitter',
+	communityRequestsTest: 'communityRequests',
+	communityContentTest: 'communityContent',
+	communityConnectionsTest: 'communityConnections'
+});
+
+export const Filter = new Entity({
+	filterActions: ['actions', FilterAction],
```
billneff79: same question: The existing name is pretty explanatory.  Why are we renaming it?

src/schema/schema.graphql
```diff
@@ -509,6 +509,197 @@ type FreeBusy {
...
+	numberComparison: String
+	size: String # s
+	index: Int
+	negative: Boolean
+}
+
+type FilterCondition {
+	allOrAny: FilterMatchCondition! # condition
+
+	addressBook: [HeaderCheckCondition] # addressBookTest
```
billneff79: Seems odd to me that it is called a "test" in the API, but we call it a "condition" in the GraphQL schema

src/schema/schema.graphql
```diff
@@ -1268,6 +1639,7 @@ type Mutation {
 	modifyAppointment(appointment: CalendarItemInput!): Boolean
 	modifyIdentity(id: ID!, attrs: IdentityAttrsInput!): String
 	modifyPrefs(prefs: PreferencesInput!): Boolean
+	modifyFilterRules(filters: [FilterInput!]): Boolean
```
billneff79: If we have FilterInput as required, is it possible to delete all of your filter rules?

src/normalize/entities.ts
silentsakky: it's parent node is already indicating that it's a condition, so removed it from each level

src/normalize/entities.ts
silentsakky: parent node already indicates that it's already part of filter, so removed it from child levels

src/schema/schema.graphql
silentsakky: in api we are having this structure

```
filterConditions 
    |-- headerTest
    |-- addressbookTest
```

which I have changed to

```
conditions
    |-- header
    |-- addressbook
```

This is done for better semantics, nothing else

src/schema/schema.graphql
silentsakky: Yes
This problem will arrive only if we had used [FilterInput]! where it will not allow empty array
but we used [FilterInput!] so it will allow empty array but will not allow empty FilterInput object

PR #16, opened 2018-05-24
PREAPPS-381: Ability to respond to shared calendar invitation in email

pl12133:
Added ShareNotification to message queries, and implemented `CreateMountpointRequest`.

Also brought back the `ZimbraErrorLink` in order to print out all GraphQL error messages to the console by default; this can avoid some hard to debug silent error messages.


src/batch-client/index.ts
```diff
@@ -193,6 +195,12 @@ export class ZimbraBatchClient {
 		});
 	};
 
+	public createMountpoint = (_options: CreateMountpointInput) =>
```
billneff79: Don't need to change it unless you are there for something else, but not sure why we are prefixing `_` for `_options` here...

PR #20, opened 2018-05-29
Hotfix: Always denormalize arguments passed to `action`

pl12133:
This is breaking moving mail messages between folders on `master`.

There are some consumers that use the raw `action` method, so we always need to be calling `denormalize` on the arguments passed it.


silentsakky:
Awesome

PR #21, opened 2018-06-01
PREAPPS-613: Implement Add/modify/delete Signature

pl12133:

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -221,6 +221,13 @@ export class ZimbraBatchClient {
 			}
 		});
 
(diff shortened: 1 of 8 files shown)


src/batch-client/index.ts
```diff
@@ -114,6 +116,13 @@ export class ZimbraBatchClient {
 		});
 	};
 
+	public addSignature = (options: SignatureInput) =>
```
billneff79: - [ ] Rename `addSignature` to `createSignature` (and put in file in appropriate alphabetical place to match `CreateSignature` request name in API

PR #23, opened 2018-06-04
Preapps 555 Choose calendar to add event to

itsmeprashant:


src/batch-client/index.ts
```diff
@@ -476,6 +497,16 @@ export class ZimbraBatchClient {
 			)
 		);
 
+	public singleRequest = (options: JsonRequestOptions) => {
+		singleJSONRequest({
+			...options,
+			sessionId: this.sessionId,
+			origin: this.origin,
+			name: options.name,
```
silentsakky: below two statements are not required

src/batch-client/index.ts
```diff
@@ -476,6 +497,16 @@ export class ZimbraBatchClient {
 			)
 		);
 
+	public singleRequest = (options: JsonRequestOptions) => {
+		singleJSONRequest({
```
silentsakky: by calling this function directly we are skipping code for notifications handling

dajay0:
Looked through. I dont have any comments.

src/batch-client/index.ts
```diff
@@ -162,20 +174,30 @@ export class ZimbraBatchClient {
...
+		appointment: CalendarItemInput
+	) =>
 		this.jsonRequest({
 			name: 'CreateAppointment',
 			body: {
 				...denormalize(CalendarItemCreateModifyRequest)(appointment)
-			}
+			},
+			accountName: accountName,
+			namespace: Namespace.Mail
```
pl12133: It looks like the namespace will default to `Namespace.Mail`, so would it be OK to remove this extra option?

itsmeprashant: Done!

PR #26, opened 2018-06-06
CircleCI Build Issues: Something wrong with Cache

pl12133:
Trying out some different cache settings to see if we can get this error to go away:

> @zimbra/api-client@1.0.11 copySchema /home/circleci/repo
> copy src/schema/schema.graphql dist
>
> sh: 1: copy: Permission denied


pl12133:
Swapped out the [`copy`](https://github.com/jonschlinkert/copy) package for [`copyfiles`](https://github.com/calvinmetcalf/copyfiles) and the permissions issue is no longer present.

I also tried `fs.copyFilesSync` which works fine, but would make us depend on Node `8.5.0` so instead changed over to `copyfiles`.

PR #29, opened 2018-06-12
PREAPPS-410: Conversation view does not refresh as soon as reply is sent.

lavegupta:
Created ConversationDetail type for conversation detail view to manage it separately.


pl12133:
We might want to update the resolver for `MailItem` so that the schema can differentiate between the different interfaces that implement `MailItem`.

https://github.com/Zimbra/zm-api-js-client/blob/111377449f6e7274bca35d2cf005b4845a979603/src/schema/schema.ts#L86-L88

I think we can use the [second argument to `__resolveType`](https://www.apollographql.com/docs/graphql-tools/resolvers#Resolver-function-signature) called `args` to tell if the request came from a `Search` or `GetConv` request.

I tried to make this work myself but it seems like the `__resolveType` function never runs, so it might not be very important... but it seems like the correct thing to do since we have a new interface implementing `MailItem`.

billneff79:
`ConversationDetail` is the exact same as the existing `Conversation` type.  This change does not appear to be necessary.  I'm going to close this PR.  Please open it again if you feel it is needed.  I will modify `zm-x-web` repo to use `Conversation` instead of `ConversationDetail`

silentsakky:
@billneff79 Actually this is required to make sure we have different caches for SearchResponse and GetConversationResponse as both will return different set of data

I will move this to ZimbraOS/zm-x-web, so we can discuss if required

silentsakky:
@itsmeprashant can we delete this branch and PR if all changes are reverted?

billneff79:
This is definitely not needed.  Going to close and delete the branch.  There is 0 reason to have the same shape of object with a different name.  Doing that likely means you need to use different fragments in your `.graphql` document queries in `zm-x-web` so you don't say you are getting data back from a query that you really aren't and thus caching null values.

Going to delete this branch so the PREAPPS-410 build in `zm-x-web` passes.

PR #30, opened 2018-06-13
Preapps 428: new schema for reply to event invite

jkester1986:


src/batch-client/index.ts
```diff
@@ -512,6 +515,14 @@ export class ZimbraBatchClient {
 			return normalized;
 		});
 
+	public sendInviteReply = (requestOptions: InviteReplyInput) =>
+		this.jsonRequest({
+			name: 'SendInviteReply',
+			body: {
+				...denormalize(InviteReply)(requestOptions)
+			}
+		}).then(res => normalize(CalendarItemHitInfo)(res));
```
pl12133: Since `normalize` returns a function, you can avoid the extra function closure here by just using the return value from `normalize`:

```js
}).then(normalize(CalendarItemHitInfo));
```

src/batch-client/index.ts
```diff
@@ -512,6 +515,14 @@ export class ZimbraBatchClient {
 			return normalized;
 		});
 
+	public sendInviteReply = (requestOptions: InviteReplyInput) =>
+		this.jsonRequest({
+			name: 'SendInviteReply',
+			body: {
+				...denormalize(InviteReply)(requestOptions)
```
pl12133: We can avoid the extra object clone here:

```js
body: denormalize(InviteReply)(requestOptions)
```

It is written like that in a few places in the code, but those should eventually be changed to not do an extra clone also.

PR #31, opened 2018-06-15
PREAPPS-629

silentsakky:
- Add login, logout and change password apis in graphql
- Make sure event emitter passes data to subscribers from BatchLink
- Fix issue where request object was trying to json parse request which had http status 502 and no data, which was passing javascript error in graphql error instead of actual error

### DIFF
--- a/src/request/index.ts
+++ b/src/request/index.ts
@@ -24,6 +24,10 @@ function soapCommandBody(options: RequestOptions) {
 }
 
 function parseJSON(response: Response): Promise<ParsedResponse> {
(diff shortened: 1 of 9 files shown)


src/request/index.ts
```diff
@@ -206,6 +206,13 @@ export function jsonRequest(
 		}),
 		headers: options.headers
 	})
+		.then(response => {
```
pl12133: This could be a littler shorter if we put the `response.ok` check into the beginning of the `parseJSON` function.

PR #35, opened 2018-06-27
PREAPPS-654: Alert user when unable to send message

lavegupta:


src/batch-client/index.ts
```diff
@@ -503,6 +505,14 @@ export class ZimbraBatchClient {
 			}
 		}).then(res => normalize(CalendarItemHitInfo)(res));
 
+	public sendMessage = (body: SendMessageInput) =>
+		this.jsonRequest({
+			name: 'SendMsg',
+			body: {
+				...denormalize(SendMessageInfo)(body)
```
jkester1986: @pl12133  said we shouldn't need to use `...` with denormalize and normalize, and is trying to get rid of their use

src/normalize/entities.ts
```diff
@@ -164,6 +174,25 @@ const commonMailItemFields = {
 	attach: ['attachments', AttachmentsInfo]
 };
 
+const SendMessageFields = new Entity({
```
jkester1986: you should be able to get a lot of these keys from commonMailItemFields (which also uses commonMessageFields) instead of declaring them all over again

jkester1986: in fact, you might be able to just modify commonMailItemFields to add anything extra that was missing

lavegupta: Thanks for pointing out this @jkester1986.
I have fixed it now please review.

PR #37, opened 2018-07-11
PREAPPS-811: grab rt from message response

jkester1986:
**NOTE** breaking change, switching `rt` to `replyType`


src/schema/generated-schema-types.ts
```diff
@@ -18,6 +18,7 @@ export interface MailItem {
 	invitations?: InviteInfo[] | null;
 	sortField?: string | null;
 	share?: ShareNotification[] | null;
+	rt?: string | null;
```
pl12133: The key `rt` is obscure enough that we should probably normalize it in `src/normalize/entities.ts`, it is called `replyType` in the documentation so could normalize it to that name?

PR #38, opened 2018-07-13
PREAPPS-742: default calendar option

jkester1986:


src/schema/schema.graphql
```diff
@@ -1755,6 +1757,7 @@ type Mutation {
 	modifyTask(task: CalendarItemInput!): Boolean
 	moveTask(inviteId: ID!, destFolderId: ID!): String
 	prefAutoAddAppointmentToCalendar(value: Boolean!): Boolean
+	prefCalendarDefaultCal(value: Int!): Int
```
silentsakky: i think instead of adding mutation for every single pref change, we should make use of modifyPrefs mutation, which can handle any number of prefs in a single request

PR #40, opened 2018-07-27
PREAPPS-856:: User should able to fetch public certificate of the recipient from Contact, GAL or LDAP. (WIP)

SantoshKShetty:
- [x] Added logic to fetch public certificates

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -381,7 +381,6 @@ export class ZimbraBatchClient {
 		);
 
 	public getSMimePublicCerts = (options: GetSMimePublicCertsOptions) =>
-		options.recipientAddr &&
 		this.jsonRequest({
 			name: 'GetSMIMEPublicCerts',
 			body: {
@@ -389,11 +388,11 @@ export class ZimbraBatchClient {
 					_content: options.store
 				},
 				email: {
-					_content: options.recipientAddr
+					_content: options.contactAddr
 				}
 			},
 			namespace: Namespace.Account
-		}).then(res => mapValuesDeep(res, coerceStringToBoolean));
+		});
 
 	public itemAction = (options: ActionOptions) =>
 		this.action(ActionType.item, options);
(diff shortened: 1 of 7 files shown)


src/batch-client/index.ts
```diff
@@ -379,6 +380,21 @@ export class ZimbraBatchClient {
...
+	public getSMimePublicCerts = (options: GetSMimePublicCertsOptions) =>
+		options.recipientAddr &&
+		this.jsonRequest({
+			name: 'GetSMIMEPublicCerts',
+			body: {
+				store: {
+					_content: options.store
+				},
+				email: {
+					_content: options.recipientAddr
```
silentsakky: shouldn't we need handle multiple email addresses here?

src/batch-client/index.ts
```diff
@@ -379,6 +380,21 @@ export class ZimbraBatchClient {
 			res => (res.search ? { folders: normalize(Folder)(res.search) } : {})
 		);
 
+	public getSMimePublicCerts = (options: GetSMimePublicCertsOptions) =>
+		options.recipientAddr &&
```
silentsakky: this should be part of your schema

src/batch-client/index.ts
```diff
@@ -379,6 +380,21 @@ export class ZimbraBatchClient {
...
+			body: {
+				store: {
+					_content: options.store
+				},
+				email: {
+					_content: options.recipientAddr
+				}
+			},
+			namespace: Namespace.Account
+		}).then(res => mapValuesDeep(res, coerceStringToBoolean));
```
silentsakky: i don't think response contains any boolean values, which needs to be converted

src/batch-client/index.ts
silentsakky: not doing it now as we are not going to send multiple addresses in new approach

src/batch-client/index.ts
silentsakky: done

src/batch-client/index.ts
silentsakky: fixed

PR #41, opened 2018-07-30
PREAPPS-822: Design for the local data archive

pl12133:
This adds `autoComplete` and `saveDraft`. ~Also creates a mechanism to queue up `jsonRequests` and then wait until we go online to fire them off.~

~This method will probably be replaced now that I'm more aware of the available tools (see https://github.com/benseitz/apollo-link-offline/issues/1#issuecomment-371678922) but it is a good start.~

The offline strategy has been replaced in #42 so it was deleted from this pull request.


src/schema/schema.graphql
```diff
@@ -1102,6 +1151,29 @@ input AttachmentInput {
...
+	id: String
+	folderId: String
+	display: String
+	first: String
+	middle: String
+	last: String
+	full: String
+	nick: String
+	company: String
+	filas: String
```
silentsakky: typo ???
fileas ???

src/schema/schema.graphql
```diff
@@ -1102,6 +1151,29 @@ input AttachmentInput {
...
 }
 
+type AutoCompleteResponse {
+	canBeCached: Boolean
+	match: [AutoCompleteMatch]
+}
+
+type AutoCompleteMatch {
+	email: String
+	type: MatchTypeGal
```
silentsakky: MatchTypeGal is a misleading name, can be changed to AutoCompleteMatchType

src/schema/schema.graphql
```diff
@@ -1792,7 +1871,8 @@ type Mutation {
 	prefOutOfOfficeFromDate(value: String!): String
 	prefOutOfOfficeReply(value: String!): String
 	prefOutOfOfficeUntilDate(value: String!): String
-	sendMessage(message: SendMessageInput!): Boolean
+	saveDraft(message: SendMessageInput!): SaveDraftResponse
+	sendMessage(message: SendMessageInput!): SendMessageResponse
```
silentsakky: Normalization is not done on SendMsg response in batch-client/index.ts

PR #42, opened 2018-08-01
PREAPPS-823: Offline message send

pl12133:
Still a lot to do here but I am opening this early because it affects #41. The link strategy is superior and the former strategy doesn't need to be reviewed.

#### TODO
 - [x] Typings for new apollo links.


index.ts
```diff
@@ -3,6 +3,8 @@ export { ZimbraBatchClient } from './src/batch-client';
 export { LocalBatchLink } from './src/apollo/local-batch-link';
 export { ZimbraErrorLink } from './src/apollo/zimbra-error-link';
 export { ZimbraInMemoryCache } from './src/apollo/zimbra-in-memory-cache';
+export { OfflineQueueLink } from './src/apollo/offline-queue-link';
```
silentsakky: instead of creating our custom implementation, can we make use of https://github.com/helfer/apollo-link-queue ?

pl12133: That works well for just queries, but it doesn't queue mutations. I took the [QueueMutationLink.js](https://github.com/iamrommel/offline-demo/blob/master/web/src/QueueMutationLink.js) that I found linked in the [`apollo-link-offline`](https://github.com/benseitz/apollo-link-offline) repo and combined it with the `apollo-link-queue` link in order to queue both queries and mutations.

I tried using the two links separately, but when one link was "closed" the other would not function correctly; so instead I combined the two into this one link.

pl12133: To correct myself above, it doesn't *persist* queued mutations. It does queue them, but not in a way that can be persisted to storage because the `forward` function cannot be serialized.

pl12133:
This became closed when #41 got closed, reopening.

PR #43, opened 2018-08-01
PREAPPS-776: graphql for password recovery

Sigafoos:


src/schema/schema.graphql
```diff
@@ -1686,6 +1707,18 @@ type Query {
 	preferences: Preferences
 	relatedContacts(email: String!): RelatedContacts
 	shareInfos(addresses: [String!]): [[ShareInfo]]
+    setRecoveryAccount(
```
jkester1986: think this needs to be under mutations

src/batch-client/index.ts
```diff
@@ -525,6 +538,21 @@ export class ZimbraBatchClient {
 			}
 		});
 
+	public setRecoveryAccount = ({
+		op,
+		recoveryAccount,
+		recoveryAccountVerificationCode
+	}: SetRecoveryAccountOptions) =>
+		this.jsonRequest({
+			name: 'SetRecoveryAccountRequest',
```
jkester1986: Take off the `Request` part of this, it gets appended on to the end automatically

src/batch-client/index.ts
```diff
@@ -474,6 +477,16 @@ export class ZimbraBatchClient {
 			namespace: Namespace.Account
 		}).then(res => mapValuesDeep(res._attrs, coerceStringToBoolean));
 
+	public recoverAccount = ({ email, op }: RecoverAccountOptions) =>
+		this.jsonRequest({
+			name: 'RecoverAccount',
+			body: {
+				channel: SetRecoveryAccountChannelType.email,
```
silentsakky: we shouldn't be hardcoding channel here, we can specify it as default parameter in query

src/batch-client/index.ts
```diff
@@ -525,6 +538,21 @@ export class ZimbraBatchClient {
...
 
+	public setRecoveryAccount = ({
+		op,
+		recoveryAccount,
+		recoveryAccountVerificationCode
+	}: SetRecoveryAccountOptions) =>
+		this.jsonRequest({
+			name: 'SetRecoveryAccountRequest',
+			body: {
+				channel: SetRecoveryAccountChannelType.email,
```
silentsakky: same as above

src/schema/schema.graphql
```diff
@@ -213,6 +213,22 @@ enum LicenseStatus {
 	ACTIVATION_GRACE_PERIOD
 }
 
+enum SetRecoveryAccountOp {
```
silentsakky: indentation looks off

src/schema/schema.ts
```diff
@@ -274,6 +277,16 @@ export function createZimbraSchema(
 								_attrs: mapValues(variables.attrs, coerceBooleanToString)
 							}
 						}
+					}),
+				setRecoveryAccount: (
+					_,
+					{ channel, op, recoveryAccount, recoveryAccountVerificationCode }
```
silentsakky: we can use SetRecoveryAccountOptions

src/schema/schema.graphql
```diff
@@ -1351,6 +1356,11 @@ type CalendarItemAlarm {
 	trigger: [CalendarItemAlarmTrigger]
 }
 
+type RecoverAccount {
+  recoveryAccount: String
+  recoveryAttemptsLeft: String
+}
```
jkester1986: looks like you have spaces instead of tabs here

src/schema/schema.graphql
```diff
@@ -1846,6 +1856,12 @@ type Query {
 	preferences: Preferences
 	relatedContacts(email: String!): RelatedContacts
 	shareInfos(addresses: [String!]): [[ShareInfo]]
+    recoverAccount(
+      op: RecoverAccountOp!
+      email: String!
+      channel: SetRecoveryAccountChannel!
+    ): RecoverAccount
```
jkester1986: also spaces instead of tabs

src/batch-client/index.ts
```diff
@@ -509,6 +536,13 @@ export class ZimbraBatchClient {
 			}
 		});
 
+	public resetPassword = (body: ResetPasswordOptions) =>
```
silentsakky: better to destructure body and get password and use it below, it makes code more cleaner and we know how many params are getting passed to api

src/schema/schema.graphql
```diff
@@ -1351,6 +1356,11 @@ type CalendarItemAlarm {
 	trigger: [CalendarItemAlarmTrigger]
 }
 
+type RecoverAccount {
+	recoveryAccount: String
+	recoveryAttemptsLeft: String
```
silentsakky: this should be Int

src/batch-client/index.ts
```diff
@@ -499,6 +512,20 @@ export class ZimbraBatchClient {
 			namespace: Namespace.Account
 		}).then(res => mapValuesDeep(res._attrs, coerceStringToBoolean));
 
+	public recoverAccount = ({
+		channel = SetRecoveryAccountChannelType.email,
```
silentsakky: channel is a mandatory param in schema.graphql, so don't think this default value will ever get used

src/batch-client/index.ts
```diff
@@ -499,6 +523,16 @@ export class ZimbraBatchClient {
...
 
+	public recoverAccount = ({ channel, email, op }: RecoverAccountOptions) =>
+		this.jsonRequest({
+			name: 'RecoverAccount',
+			body: {
+				channel,
+				email,
+				op
+			}
+		}).catch(() => ({ error: true }));
```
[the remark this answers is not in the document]
Sigafoos: If we don't then an incorrect code throws an uncaught exception.

src/schema/schema.graphql
```diff
@@ -1846,6 +1856,12 @@ type Query {
 	preferences: Preferences
 	relatedContacts(email: String!): RelatedContacts
 	shareInfos(addresses: [String!]): [[ShareInfo]]
+	recoverAccount(
```
pl12133: I think this fits better as a `mutation`, since it is un-cacheable and we're altering something on the back end. As a mutation, we can `catch` the error that happens when the mutation fails and we can remove the `catch` statement on the `ZimbraBatchClient::recoverAccount` method.

PR #45, opened 2018-08-06
PREAPPS-775: schema changes for set account recovery

jkester1986:


src/batch-client/index.ts
```diff
@@ -525,6 +526,22 @@ export class ZimbraBatchClient {
 			}
 		});
 
+	public setRecoveryAccount = ({
+		op,
+		recoveryAccount,
+		recoveryAccountVerificationCode,
+		channel
+	}: SetRecoveryAccountOptions) =>
```
pl12133: Since this entire object is passed into `jsonRequest`, we can shorten this to:

```js
	public setRecoveryAccount = (options: SetRecoveryAccountOptions) =>
		... snip ...
			body: options
```

src/schema/schema.ts
```diff
@@ -274,6 +274,16 @@ export function createZimbraSchema(
 								_attrs: mapValues(variables.attrs, coerceBooleanToString)
 							}
 						}
+					}),
+				setRecoveryAccount: (
+					_,
+					{ channel, op, recoveryAccount, recoveryAccountVerificationCode }
```
pl12133: Similar to above, this can be shortened to `variables as SetRecoveryAccountOptions` and all variables can be passed on to `client.setRecoveryAccount`.

PR #47, opened 2018-08-07
Preapps 767 - User should able to sign and send an email from Desktop application

itsmeprashant:


src/batch-client/index.ts
```diff
@@ -551,6 +551,42 @@ export class ZimbraBatchClient {
...
+			credentials: 'include'
+		}).then(response => {
+			if (response.ok) {
+				return response.text().then(result => {
+					if (!result) {
+						return null;
+					}
+
+					const [, status = '', err = undefined, json = ''] =
+						result.match(/^([^,]+),([^,]+),(.*)/) || [];
```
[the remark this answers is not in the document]
itsmeprashant: So server returns a string, which contains comma separated status, err and JSON.

src/schema/schema.graphql
```diff
@@ -1796,6 +1799,7 @@ type Mutation {
 	sendInviteReply(inviteReply: InviteReplyInput!): InviteReplyResponse
 	sendShareNotification(shareNotification: ShareNotificationInput!): Boolean
 	setMailboxMetadata(section: String, attrs: MailboxMetadataSectionAttrsInput!): Boolean
+	uploadMessage(messageString: String): String
```
silentsakky: messageString should be required parameter i guess

src/schema/schema.ts
```diff
@@ -259,6 +259,8 @@ export function createZimbraSchema(
 					client.deleteSignature(variables as NameIdInput),
 				sendMessage: (_, variables) =>
 					client.sendMessage(variables as SendMessageInput),
+				uploadMessage: (_, { messageString }) =>
```
silentsakky: variable

silentsakky: better to maintain same signature as other methods by using variables

src/batch-client/index.ts
```diff
@@ -551,6 +551,43 @@ export class ZimbraBatchClient {
...
+			},
+			credentials: 'include'
+		}).then(response => {
+			if (response.ok) {
+				return response.text().then(result => {
+					if (!result) {
+						return null;
+					}
+
+					// To parser server response like => 200,'null',[{"aid":"d93a252a-603e-4675-9e39-95cebe5a9332:b39a4b7c-9232-4228-9269-aa375bc1df67","ct":"message/rfc822","filename":"message.eml","s":7872}]
```
silentsakky: if we are only using attachmentId then can we use raw format instead of extended,raw

src/batch-client/index.ts
```diff
@@ -551,6 +551,42 @@ export class ZimbraBatchClient {
...
+		const contentType = 'message/rfc822';
+
+		return fetch(`${this.origin}/service/upload?fmt=raw`, {
+			method: 'POST',
+			body: message,
+			headers: {
+				'Content-Disposition': `${contentDisposition}; filename="${filename}"`,
+				'Content-Type': contentType
+			},
+			credentials: 'include'
```
pl12133: `credentials: 'same-origin'` is a little safer, but probably does not make a difference.

PR #48, opened 2018-08-16
Preapps 761 - As a User should able to verify the signature and authenticity of signed message

itsmeprashant:


src/batch-client/index.ts
```diff
@@ -254,6 +254,28 @@ export class ZimbraBatchClient {
 			body: options
 		});
 
+	public downloadMessage = ({ id }: any) => {
+		return fetch(`${this.origin}/service/home/~/?auth=co&id=${id}`, {
+			method: 'GET',
```
pl12133: `GET` is default and can be omitted.

src/batch-client/index.ts
```diff
@@ -254,6 +254,28 @@ export class ZimbraBatchClient {
...
+				return response.text().then(content => {
+					if (!content) {
+						return undefined;
+					}
+					return {
+						id,
+						content
+					};
+				});
+			}
```
pl12133: Can this `return Promise.reject` with some information instead of with an implicit `return Promise.resolve(undefined)`?

src/batch-client/index.ts
```diff
@@ -84,8 +84,14 @@ function normalizeMessage(
 	message: { [key: string]: any },
 	zimbraOrigin?: string
 ) {
+	const normalizedMessage = normalize(MessageInfo)(message);
+	normalizedMessage.attributes = mapValuesDeep(
```
silentsakky: please check if attributes are present then only do this processing

src/batch-client/index.ts
itsmeprashant: since it is returning then of .text(), it doesn't allow return Promise.reject.

src/batch-client/index.ts
itsmeprashant: Done!

src/schema/schema.graphql
```diff
@@ -1804,6 +1815,8 @@ type Query {
 		folders: String,
 		includeGal: Boolean
 	): AutoCompleteResponse
+	downloadMessage(id: ID!): SMimeMessage
+	folder(id: ID!): Folder
```
silentsakky: is this needed?

itsmeprashant: no.. merge issue i guess .. removing folder ..

PR #50, opened 2018-08-23
PREAPPS-824: Offline email editing

pl12133:
Refactor `offline-queue-link` to maintain one queue instead of two.

Introduces **queue control** features for the offline-queue-link. This lets us associate unique identifiers with operations, and use that identifier to have new operations cancel old operations.

See the new `offline-queue-link` README for examples.

### DIFF
--- a/src/apollo/types.ts
+++ b/src/apollo/types.ts
@@ -1,5 +1,7 @@
+import { FetchResult, NextLink, Operation } from 'apollo-link';
 import { BatchLink } from 'apollo-link-batch';
 import { GraphQLSchema } from 'graphql/type';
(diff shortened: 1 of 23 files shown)


src/apollo/types.ts
```diff
@@ -1,5 +1,7 @@
+import { FetchResult, NextLink, Operation } from 'apollo-link';
 import { BatchLink } from 'apollo-link-batch';
 import { GraphQLSchema } from 'graphql/type';
+import { Observer } from 'zen-observable-ts';
```
silentsakky: although this is included by apollo-link in node_modules,
but should we add this as an explicit dependency in package.json?

PR #51, opened 2018-08-27
PREAPPS-759

SantoshKShetty:
Committing this file due to below reason
     - PREAPPS-759 don't have any api-js change. But,
     - its depenedent upon electron_builder which is not yet pushed on master branch.
     - Since build looks for PREAPPS-759 api-js branch and since it doesn't exists, looks for master branch and fails since it require some changes of api-js.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,80 +5,54 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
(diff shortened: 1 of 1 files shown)


silentsakky:
You don't need to create PR if you don't have any changes in branch

PR #53, opened 2018-08-30
PREAPPS-1038: Add `tokenType` option to `AuthRequest`

pl12133:
Also pass on response of `AuthRequest` instead of resolving with boolean.

### DIFF
--- a/src/schema/generated-schema-types.ts
+++ b/src/schema/generated-schema-types.ts
@@ -935,7 +935,7 @@ export interface Mutation {
 	folderAction?: boolean | null;
 	itemAction?: boolean | null;
 	logout?: boolean | null;
-	login?: boolean | null;
+	login?: AuthResponse | null;
 	messageAction?: boolean | null;
 	modifyExternalAccount?: string | null;
 	modifyAppointment?: boolean | null;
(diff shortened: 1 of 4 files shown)


src/schema/generated-schema-types.ts
```diff
@@ -977,6 +977,26 @@ export interface NameId {
 	name?: string | null;
 }
 
+export interface AuthResponse {
+	authToken?: AuthToken[] | null;
```
silentsakky: usually auth token is stored as cookie only, we should not store it somewhere else

pl12133: I think you are right and this is actually already a problem, see https://github.com/apollographql/apollo-client/issues/3592.

It looks like we should likely be using `fetchPolicy: 'no-cache'` for the `AuthRequest` endpoint to workaround the issue.

plobbes: Thanks @silentsakky and @pl12133.  I agree we should not be storing/caching the auth tokens here.

silentsakky: or maybe if authtoken is not used anywhere in the application then remove it from schema? that way we will not be storing in cache at all

PR #55, opened 2018-09-04
PREAPPS-1067: make InviteType optional

jkester1986:


src/schema/schema.graphql
```diff
@@ -415,7 +415,7 @@ enum InviteType {
 }
 
 type InviteInfo {
-	type: InviteType!
+	type: InviteType #though type is a required value on the response, outdated invites come back as empty InviteInfo objects
```
pl12133: So the docs say this is required but it can actually come back as `null`? That may be a bug in the docs.

jkester1986: yeah, SOAP will return `inv: [{}]` for outdated events

PR #57, opened 2018-09-06
PREAPPS-803: Saved Search context menu

lavegupta:


src/batch-client/index.ts
```diff
@@ -460,6 +461,12 @@ export class ZimbraBatchClient {
 			}
 		});
 
+	public modifySearchQuery = (options: SearchFolderInput) =>
```
pl12133: Can we rename this to `modifySearchFolder` to match the name of the endpoint being used?

jkester1986: agreed

PR #64, opened 2018-09-13
Preapps 898

Sigafoos:


src/batch-client/index.ts
```diff
@@ -566,6 +567,92 @@ export class ZimbraBatchClient {
...
+		resultMode,
+		sortBy,
+		timezone,
+		types,
+		wantContent
+	}: SearchConversationOptions) => {
+		let body: any;
+		body = {
+			cid: conversationId,
+			...{
```
silentsakky: do we need this line, as it creates object and destructures it, we can remove it entirely

src/batch-client/index.ts
```diff
@@ -566,6 +567,92 @@ export class ZimbraBatchClient {
 			return normalized;
 		});
 
+	public searchConversation = ({
```
silentsakky:
```
public searchConversation = ({
	conversationId,
	query,
	header,
	timezone,
	locale,
	cursor,
	...options
}: SearchConversationOptions) => {
	let body: any;
	body = {
		cid: conversationId,
		...{
			_content: {
				...(query && { query }),
				...(header && { header }),
				...(timezone && { tz: timezone }),
				...(locale && { locale }),
				...(cursor && { cursor })
			}
		}
	};

	each(options, (value, key) => {
		if (value) {
			if (key === 'allowableTaskStatus' || key === 'types') {
				body[key] = value.join(',');
			} else {
				body[key] = value;
			}
		}
	});

	return this.jsonRequest({
		name: 'SearchConv',
		body: body
	}).then(res => {
		const normalized = normalize(SearchResponse)(res);
		if (get(normalized, 'conversations.0.messages')) {
			normalized.conversations[0].messages = normalized.conversations[0].messages.map(
				(m: any) => normalizeMessage(m, this.origin)
			);
		}
		return normalized;
	});
};
```

src/schema/schema.graphql
```diff
@@ -227,6 +227,43 @@ enum LicenseStatus {
 	ACTIVATION_GRACE_PERIOD
 }
 
+enum SearchConversationTaskStatus {
```
silentsakky: this is used in tasks as well, we should generically name it as TaskStatus instead of SearchConversationTaskStatus

src/schema/schema.graphql
```diff
@@ -227,6 +227,43 @@ enum LicenseStatus {
...
+	NEED
+	WAITING
+}
+
+enum SearchConversationResultMode {
+	IDS
+	NORMAL
+}
+
+enum SearchConversationSortBy {
```
silentsakky: isn't this same as "enum SortBy" ?

src/batch-client/index.ts
Sigafoos: `each` isn't a function, and I can't find a reference to it (in lodash or something similar) in this repo or the zm-x-web one. I tried `Array.prototype.slice.call(options)` (which I found in zm-x-web) but that produces an empty array.

silentsakky: https://lodash.com/docs/4.17.10#forEach, each is alias of forEach

import forEach from 'lodash-es/forEach';

src/batch-client/index.ts
```diff
@@ -566,6 +568,52 @@ export class ZimbraBatchClient {
...
+				...(cursor && { cursor })
+			}
+		};
+
+		// conditionally include the rest of the parameters
+		forEach(options, (value: any, key: string) => {
+			if (value) {
+				if (key === 'allowableTaskStatus' || key === 'types') {
+					body[key] = value.join(',');
+				} else {
```
jkester1986: else on next line

Sigafoos: The linter requires same line:

`prettier: Replace ⏎↹↹↹↹ with ·`

jkester1986: of course it does. Because why be consistent with our zm-x-web repo :P

src/schema/schema.graphql
```diff
@@ -2004,6 +2043,38 @@ type Query {
 		sortBy: SortBy,
 		types: SearchType
 	): SearchResponse
+	searchConversation(
+		allowableTaskStatus: [TaskStatus],
```
itsmeprashant: can we define `input` for these params?

pl12133:
Will be re-opened when we have more time to re-prioritize PREAPPS-898.

PR #66, opened 2018-09-19
Preapps 668

panditsavitags:


preapps668.patch
```diff
@@ -0,0 +1,561 @@
+diff --git a/package-lock.json b/package-lock.json
```
silentsakky: please remove this file

src/normalize/entities.ts
```diff
@@ -102,6 +102,14 @@ const commonInviteFields = {
 	ridZ: 'utcRecurrenceId'
 };
 
+export const ModifyAppointmentRes = new Entity({
```
silentsakky: no need to create this, we already have CalendarItemCreateModifyRequest

PR #69, opened 2018-09-20
PREAPPS-1151: Implement JWT support into next-gen-client

pl12133:
Add `jwtToken` property for using a JWT for auth.
Add `setJwtToken` method for setting `jwtToken`.


src/batch-client/index.ts
```diff
@@ -675,6 +676,7 @@ export class ZimbraBatchClient {
 	private batchDataHandler = (requests: Array<RequestOptions>) =>
 		batchJsonRequest({
 			requests,
+			jwtToken: this.jwtToken,
```
silentsakky: should we need to add same code in dataHandler function as well?

PR #73, opened 2018-09-28
Reminders Queries

Faultless:
This PR adds a Query and 2 Mutations in order to replace direct API calls with GraphQL calls on the client.

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -302,6 +302,15 @@ export class ZimbraBatchClient {
 			body: options
 		});
 
+	public dismissCalendarItem = (appointment: any, task: any) =>
+		this.jsonRequest({
+			name: 'DismissCalendarItemAlarm',
+			body: {
+				appt: appointment,
+				task
+			}
+		}).then(res => !!res);
+
 	public downloadMessage = ({ id }: any) => {
 		return fetch(`${this.origin}/service/home/~/?auth=co&id=${id}`, {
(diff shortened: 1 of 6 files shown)


src/batch-client/index.ts
```diff
@@ -301,6 +301,15 @@ export class ZimbraBatchClient {
...
 		});
 
+	public dismissCalendarItem = (appointment: any, task: any) =>
+		this.jsonRequest({
+			name: 'DismissCalendarItemAlarm',
+			body: {
+				appt: appointment,
+				task
+			}
+		}).then(res => !!res);
```
pl12133: To cast to a boolean without the extra function declaration you can use the `Boolean` constructor itself: `.then(Boolean)`.

src/batch-client/index.ts
```diff
@@ -635,6 +644,15 @@ export class ZimbraBatchClient {
 			return normalized;
 		});
 
+	public searchAppointments = (options: SearchOptions) =>
```
pl12133: Does the other `search` function not work for appointments? This one looks very similar.

Faultless:
I will need access to merge this, or perhaps someone with write access can do it for me.

pl12133:
@Faultless sure I can merge and publish this.

pl12133:
OK these changes are release as `@zimbra/api-client@7.2.0`.

Faultless:
thank you!

PR #74, opened 2018-10-01
PREAPPS-1199: Adding external account broken, failed due to "t.zimbra.account.addExternal is not a function" error

silentsakky:


jkester1986:
code looks good, got a merge conflict but should be a straightforward fix

PR #77, opened 2018-10-04
ZX-169 - Ability to upload photo for contacts [WIP]

celinakhalife:


src/normalize/entities.ts
```diff
@@ -432,7 +432,8 @@ export const GetFolderRequest = new Entity({
 
 const ContactInputAttributes = new Entity({
 	n: 'name',
-	_content: 'content'
+	_content: 'content',
+	aid: 'aid'
```
silentsakky: no need to specify entity if both names are same, like aid

src/batch-client/index.ts
```diff
@@ -524,12 +531,19 @@ export class ZimbraBatchClient {
...
 		const modifiedAttrs = <Object[]>[];
 
-		forEach(attributes, (val, key) =>
-			modifiedAttrs.push({
-				name: key,
-				content: val
-			})
-		);
+		forEach(attributes, (val, key) =>{
+			if (key === 'image') {
```
pl12133: This can be a little shorter:

```js
modifiedAttrs.push({
	name: key,
	[key === 'image' ? 'aid' : 'content']: val
})
```

PR #81, opened 2018-10-12
PREAPPS-1056: Support Locations (room resources) in events

avadhutprabhudesai:


src/batch-client/index.ts
```diff
@@ -184,6 +186,15 @@ export class ZimbraBatchClient {
 			body: denormalize(AutoCompleteEntity)(options)
 		}).then(normalize(AutoCompleteResponseEntity));
 
+	public autoCompleteGAL = (options: AutoCompleteGALOptions) =>
+		this.jsonRequest({
+			name: 'AutoCompleteGal',
+			namespace: Namespace.Account,
+			body: options
+		}).then(res => {
```
silentsakky: `}).then(res => normalize(AutoCompleteGALResponseEntity)(res));`

src/schema/schema.graphql
```diff
@@ -1347,6 +1347,21 @@ type AutoCompleteResponse {
 	match: [AutoCompleteMatch]
 }
 
+type AutoCompleteGALResponse {
+	contacts: [AutoCompleteCN]
```
silentsakky: instead of AutoCompleteCN can we use Contact? as per my understanding AutoCompleteGal always returns contact objects

src/schema/schema.graphql
```diff
@@ -2002,6 +2019,12 @@ type Query {
 		folders: String,
 		includeGal: Boolean
 	): AutoCompleteResponse
+	autoCompleteGAL(
+		limit: Int
+		name: String
```
silentsakky: name is required input

src/batch-client/types.ts
```diff
@@ -21,6 +21,12 @@ export interface AutoCompleteOptions {
 	needExp?: boolean;
 	type?: GalSearchType;
 }
+export interface AutoCompleteGALOptions {
+	limit: number;
```
silentsakky: name is only required input, all others should be optional

src/batch-client/types.ts
```diff
@@ -21,6 +21,12 @@ export interface AutoCompleteOptions {
...
+	name: string;
+	needExp: number;
```
silentsakky: Boolean

src/batch-client/types.ts
```diff
@@ -21,6 +21,12 @@ export interface AutoCompleteOptions {
...
+	type: string;
```
silentsakky: this should be an enum, with below possible values
`all|account|resource|group`

src/schema/schema.graphql
```diff
@@ -2002,6 +2019,12 @@ type Query {
...
+		type: String
```
silentsakky: this should be an enum, with below possible values
`all|account|resource|group`

src/schema/schema.graphql
```diff
@@ -2002,6 +2019,12 @@ type Query {
...
+		needExp: Int
```
silentsakky: Boolean

src/batch-client/types.ts
avadhutprabhudesai: Changed to enum

src/schema/schema.graphql
avadhutprabhudesai: I have changed this to an enum but I do not know why it isn't reflected.

src/batch-client/index.ts
```diff
@@ -9,6 +9,7 @@ import { denormalize, normalize } from '../normalize';
 import {
 	ActionOptions as ActionOptionsEntity,
 	AutoComplete as AutoCompleteEntity,
+	AutoCompleteGALResponse as AutoCompleteGALResponseEntity,
```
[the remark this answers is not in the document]
avadhutprabhudesai: Alias removed

src/schema/schema.graphql
avadhutprabhudesai: Name changed as a required attribute.

src/schema/schema.graphql
avadhutprabhudesai: ZimbraCalResType attribute is not available on type Contact. And this attribute is being used to distinguish users from resources.

src/batch-client/types.ts
avadhutprabhudesai: Changed and pushed

src/schema/schema.graphql
avadhutprabhudesai: In api docs, this attribute has value 0 | 1. Should it be changed to Boolean?

src/schema/schema.graphql
silentsakky: you can add it in Contact type, contacts can have any number of custom fields

src/schema/schema.graphql
silentsakky: Yes, api will never specify boolean as it is soap api, it will always use 0/1

src/batch-client/types.ts
avadhutprabhudesai: Changed and pushed

src/schema/schema.graphql
avadhutprabhudesai: Boolean used instead of Int and pushed

PR #85, opened 2018-10-19
PREAPPS-1101 : Add GAL as a source in the Contact Chooser

panditsavitags:


src/schema/schema.graphql
```diff
@@ -2122,6 +2129,19 @@ type Query {
 		sortBy: SortBy,
 		types: SearchType
 	): SearchResponse
+
+	searchGal(
+		needIsOwner: Boolean,
+		needIsMember: NeedIsMemberType,
+		type: String,
```
silentsakky: type should be enum

src/batch-client/index.ts
```diff
@@ -730,6 +730,13 @@ export class ZimbraBatchClient {
 			return normalized;
 		});
 
+	public searchGal = (options: SearchOptions) =>
+		this.jsonRequest({
+			name: 'SearchGal',
+			body: options,
+			namespace: Namespace.Account
+		}).then(res => normalize(SearchResponse)(res));
```
pl12133: Since `normalize` returns a function, this can be shortened to:

```js
}).then(normalize(SearchResponse));
```

PR #87, opened 2018-10-25
Preapps-716: Add action to Delete events

anukritiJagatramka:


src/normalize/entities.ts
```diff
@@ -212,6 +212,18 @@ export const CalendarItemCreateModifyRequest = new Entity({
...
 });
 
+export const InstanceDate = new Entity({
+	d: 'date'
+});
+
+export const CalendarItemDeleteRequest = new Entity({
+	inst: ['instanceDate', InstanceDate],
+	id: 'inviteId',
+	comp: 'componentNumber',
```
pl12133: Can we call this `componentNum` since there are already a few of those in this file?

src/normalize/entities.ts
```diff
@@ -410,7 +422,8 @@ export const Filter = new Entity({
 export const InviteReply = new Entity({
 	compNum: 'componentNum',
 	m: ['message', MessageInfo],
-	rt: 'replyType'
+	rt: 'replyType',
+	exceptId: ['specificInvite', InstanceDate]
```
pl12133: This is named `exceptId` on another `Invite` entity in the file, can we change this to be `exceptId` to be consistent?

src/normalize/entities.ts
anukritiJagatramka: Resolved above review comment.

src/normalize/entities.ts
anukritiJagatramka: Resolved above review comment.

src/schema/schema.ts
```diff
@@ -196,8 +197,8 @@ export function createZimbraSchema(
 					),
 				createMountpoint: (_, variables) =>
 					client.createMountpoint(variables as CreateMountpointInput),
-				deleteAppointment: (_, { inviteId }, { zimbra }) =>
-					zimbra.appointments.delete({ inviteId }),
```
silentsakky: please remove previous method declaration from zimbra-client/index.js

anukritiJagatramka: @silentsakky 
If my understanding is correct I need to remove the delete method from api.appointments, from zm-x-web\src\lib\zimbra-client\index.js.
Please confirm

silentsakky: correct

anukritiJagatramka: Fixed and pushed.

PR #90, opened 2018-10-26
ZX-320 Return the ID of a created appointment from createAppointment

silentsakky:


michaelmedellin:
@pl12133 how can we move forward with merging this PR?

PR #93, opened 2018-10-30
Preapps 1179 - First Draft autosave collapses CC/Bcc, moves cursor focus

itsmeprashant:


src/schema/schema.graphql
```diff
@@ -1406,9 +1406,14 @@ input CalendarItemMessageInput {
...
 }
 
+input NewAttachmentInput {
+	attachmentId: ID
+}
+
 input SendMessageInput {
 	id: ID #id
 	origId: ID #origid
+	attach: [NewAttachmentInput]
```
silentsakky: we already have AttachmentInput for this

src/utils/normalize-mime-parts.ts
```diff
@@ -89,6 +89,12 @@ export function normalizeMimeParts(
 				part.contentType !== 'application/pkcs7-mime' &&
 					part.contentType !== 'application/pkcs7-signature' &&
 					(acc[mode] || (acc[mode] = [])).push(processAttachment(part));
+
+				message.attributes = message.attributes || {};
+				message.attributes.isEncrypted =
+					part.contentType === 'application/pkcs7-mime';
+				message.attributes.isSigned =
+					part.contentType === 'application/pkcs7-signature';
```
silentsakky: also add application/x-pkcs7-signature check

src/utils/normalize-mime-parts.ts
```diff
@@ -89,6 +89,12 @@ export function normalizeMimeParts(
 				part.contentType !== 'application/pkcs7-mime' &&
 					part.contentType !== 'application/pkcs7-signature' &&
 					(acc[mode] || (acc[mode] = [])).push(processAttachment(part));
+
+				message.attributes = message.attributes || {};
+				message.attributes.isEncrypted =
+					part.contentType === 'application/pkcs7-mime';
```
silentsakky: do we need check for multipart/signed also?

itsmeprashant: Yes, we will add that when we gonna support Outlook/Thunderbird client.

src/utils/normalize-mime-parts.ts
itsmeprashant: sure.

src/schema/schema.graphql
itsmeprashant: That is different and used for other purpose. We can't use that.

silentsakky: I think it's same, it has extra key which can be ignored

PR #94, opened 2018-10-31
 ZX-168 - Mail Search Right sidebar metadata: Analytics, Recent Photos, Recent Documents and related Contacts

celinakhalife:


celinakhalife:
hey @pl12133, kindly note that I pushed a commit after your approval.
If all is good can we merge this PR?

PR #107, opened 2018-11-29
ZX-402: fix contact notification handler to read and write correctly to cache

aschneit:


silentsakky:
This PR fixes issues reported in PREAPPS-1592

PR #110, opened 2018-12-13
PREAPPS-1083 Invitation updates should show current response with option to update

damini-vashishtha:


src/normalize/entities.ts
```diff
@@ -119,6 +124,12 @@ const InviteComponent = new Entity({
 	seq: 'sequence'
 });
 
+const InviteReplies = new Entity({
+	...commonMessageFields,
```
silentsakky: InviteReplies has only one key, so I guess these two lines are not required

damini-vashishtha: Done

PR #112, opened 2018-12-14
PREAPPS-898: Expand all unread messages in a conversation.

pl12133:


src/batch-client/index.ts
```diff
@@ -777,6 +778,47 @@ export class ZimbraBatchClient {
...
+		conversationId,
+		query,
+		header,
+		timezone,
+		locale,
+		cursor,
+		...options
+	}: SearchConversationOptions) => {
+		let body: any = {
+			cid: conversationId,
```
silentsakky: should this conversion happen using normalize@index.js#denormalize method instead of hardcoding here?

src/normalize/entities.ts
```diff
@@ -187,21 +187,24 @@ export const SendMessageInfo = new Entity({
...
 	n: 'numMessages',
 	m: ['messages', MessageInfo],
 	u: 'unread'
 });
 
-export const SearchConversation = new Entity({
+export const ConversationHitInfo = new Entity({
 	...commonMailItemFields,
 	n: 'numMessages',
-	m: ['messagesMetaData', MessageInfo],
```
pl12133: I've renamed this property back to messages, I'm not sure I totally understood the change from https://github.com/Zimbra/zm-api-js-client/pull/117 but it looked like the conflicting types were overwriting each other (`SearchResponse` and `GetConvResponse`). This shouldn't happen anymore since there are now three distinct conversation types - `ConversationInfo`, `ConversationHitInfo`, and `NestedSearchConversation` - so they should not overwrite one another.

PR #114, opened 2018-12-18
Simplify `audit-ci` expression

quinnturner:
Only the `moderate` flag is required to catch `moderate`, `high`, and `critical` vulnerabilities.


silentsakky:
Thanks for the contribution, I missed this simple logic :)

PR #117, opened 2018-12-24
PREAPPS:1689- Moving conversation into local folders.

anukritiJagatramka:


src/normalize/entities.ts
```diff
@@ -198,6 +198,13 @@ export const Conversation = new Entity({
 	u: 'unread'
 });
 
+export const SearchConversation = new Entity({
+	...commonMailItemFields,
+	n: 'numMessages',
+	m: ['messagesMetaData', MessageInfo],
```
pl12133: Why is `Conversation` separated from `SearchConversation` and `messages` renamed to `messagesMetaData`? This is a breaking change and it conflicts with PREAPPS-898.

PR #118, opened 2018-12-24
PREAPPS-1631 Clicking on Local Folder should show the list of emails

itsmeprashant:


src/apollo/zimbra-in-memory-cache.ts
```diff
@@ -55,23 +55,29 @@ function createFragmentMatcher(fragmentMatcherFactory = Object) {
...
+};
+
 /**
  * Provide a light wrapper over Apollo's inmemory cache with
  * special optimizations for identifying Zimbra object types via
  * `dataIdFromObject`.
  */
 export class ZimbraInMemoryCache extends InMemoryCache {
-	constructor(config: ApolloReducerConfig = {}) {
+	public name: string;
```
silentsakky: should we make this private so no one can change it post initialization

PR #120, opened 2018-12-31
PREAPPS-1055: Forward invite message as an event instead of as a message

avadhutprabhudesai:


src/normalize/entities.ts
```diff
@@ -306,6 +306,11 @@ Folder.addMapping({
 });
 export { Folder };
 
+export const ForwardAppointmentInviteInfo = new Entity({
+	id: 'id',
```
SantoshKShetty: Is this line required ?

khalidkadiri:
This is an obsolete PR. Hence closing it.

PR #121, opened 2019-01-02
added ciFolder to commonInviteFields entity

panditsavitags:


src/normalize/entities.ts
```diff
@@ -93,6 +93,7 @@ const commonMessageFields = {
 const commonInviteFields = {
 	compNum: 'componentNum',
 	calItemId: 'calendarItemId',
+	ciFolder: 'ciFolder',
```
silentsakky: if key and value is same then it should not appear in entities.ts file

PR #122, opened 2019-01-04
PREAPPS-1316: Upgrade to Webpack 4 

pl12133:
This upgrades our GraphQL stack to `14` and some related types.


pl12133:
For some reason the `graphql` upgrade was entering an infinite loop when parsing the Array-like objects that come from `zm-x-web::src/lib/zimbra-client`. That stuff is deprecated anyway so I was (finally) able to remove the dependency on the deprecated API client in https://github.com/ZimbraOS/zm-x-web/pull/797/commits/f0938cfccd436eeaff0684173dc8ca7a0eb45371

PR #128, opened 2019-02-01
PREAPPS-1790: New contact list gets inserted at top then automatically moves, confusing the user

madhavdhuppe:


src/schema/notifications.ts
```diff
@@ -98,13 +119,18 @@ export class ZimbraNotifications {
 				const id = findDataId(this.cache, '$ROOT_QUERY.search', dataId =>
 					r.test(dataId)
 				);
+				const sortByValue = getVariablesFromDataId(id)
+					? getVariablesFromDataId(id).sortBy
```
SantoshKShetty: const { sortBy } = getVariablesFromDataId(id)

src/schema/notifications.ts
```diff
@@ -127,7 +153,9 @@ export class ZimbraNotifications {
 					}
 				});
 				searchResponse[query].contacts = uniqBy(
-					[item].concat(searchResponse[query].contacts),
+					[item]
+						.concat(searchResponse[query].contacts)
+						.sort((a, b) => doSort(a, b, sortByValue)),
```
SantoshKShetty: It's better to insert the item instead of sorting every time when a new contact is added because, `searchResponse[query].contacts` is previous cached data which we can believe already sorted

src/schema/notifications.ts
madhavdhuppe: while deleting contact groups from we get undefined id so need to set those conditions

src/schema/notifications.ts
```diff
@@ -126,10 +156,17 @@ export class ZimbraNotifications {
 						...item
 					}
 				});
+				if (sortBy) {
```
silentsakky: if there is no sortyBy attribute then we will not concat new contact with search response, which is a bug

madhavdhuppe: as per my understanding when user deletes the contactgroup from list then in that case sortBy value is coming undefined.
so, that's why i put check here
while inserting into list we will always get the sortBy value

madhavdhuppe: is there any possibility while inserting we will get no sortBy value ?

if yes, then I will remove the check.

src/schema/notifications.ts
```diff
@@ -41,6 +41,34 @@ function findDataId(
...
+				array.splice(i, 0, item);
+				return array;
+			} else if (i === array.length - 1) {
+				array.push(item);
+				return array;
+			}
+		}
+	} else {
+		array = [item].concat(array);
+		return array;
```
sachinpatel88: Can we have single `return array` in this method?

madhavdhuppe: done

src/schema/notifications.ts
silentsakky: don't just look at current scenarios, in future someone introduces a call without sortBy value then there will be a bug introduced unknowingly

src/schema/notifications.ts
```diff
@@ -98,13 +126,15 @@ export class ZimbraNotifications {
 				const id = findDataId(this.cache, '$ROOT_QUERY.search', dataId =>
 					r.test(dataId)
 				);
+				const { sortBy }: any = getVariablesFromDataId(id) || {};
```
silentsakky: sortBy always contains strings, any reason to give {} as default value?

src/schema/notifications.ts
```diff
@@ -41,6 +41,34 @@ function findDataId(
 	)[0];
 }
 
+function addNewContactGroupToList(array: any, item: any, sortBy: any) {
+	if (sortBy === 'nameAsc' && array.length > 0) {
```
silentsakky: array is a common javascript keyword, please don't use as variable name, can confuse any new developer

src/schema/notifications.ts
madhavdhuppe: ok will remove the check

src/schema/notifications.ts
```diff
@@ -41,6 +41,34 @@ function findDataId(
...
+		for (let i = 0; i < array.length; i++) {
```
itsmeprashant: We can use lodash/reduce for the following operation.

src/schema/notifications.ts
```diff
@@ -41,6 +41,34 @@ function findDataId(
...
+			if (comparedResult === -1) {
+				array.splice(i, 0, item);
+				break;
+			} else if (i === array.length - 1) {
+				array.push(item);
+				break;
+			}
+		}
+	} else {
+		array = [item].concat(array);
```
silentsakky: item is an object, so better would be to use array.push instead of concat

src/schema/notifications.ts
madhavdhuppe: getVariablesFromDataId(id) this function returns the json object of provide id and from that json we are destructing sortBy while adding item into list.
But while deleting item from list, we get don't get any "id" so function will return 'undefined' so to handle that i'm doing this. as per discussed with santosh

src/schema/notifications.ts
silentsakky: there could a simpler approach
```
const index = array.findIndex(cnt => item.fileAsStr.localeCompare(cnt.fileAsStr. undefined, { sensitivity: 'base' }) === -1);
if (index !== -1) {
	array.splice(index + 1, 0, item);
} else {
	array.push(item);
}
```

src/schema/notifications.ts
madhavdhuppe: Updated

src/schema/notifications.ts
madhavdhuppe: updated

madhavdhuppe: can you please explain on this

src/schema/notifications.ts
```diff
@@ -41,6 +41,31 @@ function findDataId(
...
+					sensitivity: 'base'
+				}) === -1
+		);
+		if (index !== -1) {
+			itemList.splice(index, 0, item);
+		} else {
+			itemList.push(item);
+		}
+	} else {
+		itemList = [item].concat(itemList);
```
silentsakky: push

madhavdhuppe: Here, Initial approach was like this so kept same,

when the list is empty the 'searchResponse[query].contacts' is not coming as array so it's not getting pushed as expected to handle that it has written like that.

madhavdhuppe: it gives problem when the there is no list and we are adding a new item.

PR #129, opened 2019-02-05
PREAPPS-848: MOVE email items FROM local store into other folders

lavegupta:


src/schema/schema.graphql
```diff
@@ -2063,6 +2063,10 @@ input DismissInput {
 	dismissedAt: Float!
 }
 
+input AddMsgInput {
+	folderId: ID!
+	attachmentId: String!
```
SantoshKShetty: Would it be good if we use `ID` for `attachmentId` instead of `String`. I saw another place where `String` is being used, though `ID` is also kind of `String` here.

silentsakky: yes it should be ID

lavegupta: Fixed

src/normalize/entities.ts
```diff
@@ -483,3 +483,12 @@ export const ContactInputRequest = new Entity({
 	a: ['attributes', ContactInputAttributes],
 	m: 'memberOps'
 });
+
+const AddMsgAttributes = new Entity({
+	aid: 'attachmentId',
+	l: 'folderId'
```
itsmeprashant: AddMsg API supports content in the request. So instead or uploading MIME first and then passing aid inside this API, we can directly pass content and folderId.

lavegupta: Fixed, Thanks for pointing out this.

src/schema/schema.graphql
```diff
@@ -2063,17 +2063,25 @@ input DismissInput {
 	dismissedAt: Float!
 }
 
+input StringInput {
```
silentsakky: StringContentInput

lavegupta: Done

PR #133, opened 2019-02-11
PREAPPS 1173: Display invitee's local meeting time on event invitations when sender is in a different time zone

panditsavitags:


src/normalize/entities.ts
```diff
@@ -98,6 +99,7 @@ const commonMessageFields = {
 const commonInviteFields = {
 	compNum: 'componentNum',
 	calItemId: 'calendarItemId',
+	ciFolder: 'ciFolder',
```
pl12133: If this is not being renamed it can be left out.

panditsavitags: removed

PR #142, opened 2019-02-25
Preapps 1629 Add second response for Enterprise Out of Office

damini-vashishtha:


src/schema/schema.graphql
```diff
@@ -2353,8 +2359,11 @@ type Mutation {
 	moveTask(inviteId: ID!, destFolderId: ID!): String
 	prefEnableOutOfOfficeAlertOnLogin(value: Boolean!): Boolean
 	prefEnableOutOfOfficeReply(value: Boolean!): Boolean
+	prefEnableOutOfOfficeExternalReply(value: Boolean!): Boolean
```
silentsakky: instead of creating separate mutation for individual property changes use modifyPrefs

PR #146, opened 2019-02-28
PREAPPS-1889: Delegation | Adding a delegate

pl12133:
Add `GetRights`, `GrantRights`, and `RevokeRights`.

These endpoints gave me some confusion because the SOAP representation is not able to translate directly to JSON, but this has been addressed in the docs at https://wiki.zimbra.com/wiki/Json_format_to_represent_soap#Example_.232:

> ```xml
> <users>
>    <user name="fernando"/>
>    <user name="frank"/>
></users>
>```
>
>If an element node has different siblings of the same node name then the object is translated into an >array of object:
>
>```js
>{users: {user: [{name: "fernando"}, {name: "frank"}]}}
>```


package.json
```diff
@@ -42,14 +42,14 @@
   "repository": "Zimbra/zm-api-js-client",
   "license": "BSD-3-Clause",
   "dependencies": {
-    "apollo-cache-inmemory": "^1.3.12",
-    "apollo-link": "^1.2.1",
-    "apollo-link-batch": "^1.1.2",
-    "apollo-link-error": "^1.0.7",
+    "apollo-cache-inmemory": "^1.5.1",
```
silentsakky: these changes are already in develop, can we merge develop in this branch?

PR #147, opened 2019-03-04
PREAPPS-1531: Contacts | Restore contact does not work in first attempt

madhavdhuppe:


src/schema/notifications.ts
```diff
@@ -42,7 +42,7 @@ function findDataId(
 }
 
 function addNewItemToList(itemList: any, item: any, sortBy: any) {
-	if (sortBy === 'nameAsc' && itemList.length > 0) {
+	if (sortBy === 'nameAsc' && item.fileAsStr && itemList.length > 0) {
```
silentsakky: curious to know, in which case we are not getting fileAsStr ?

madhavdhuppe: We are not getting "fileAsStr" when we are trying to add "contact" not "ContactGroup" into the list.
While adding contact we are only getting "id and folderId" params,

yashrajpchavda: Okay

PR #151, opened 2019-03-14
PREAPPS-1497: HTML email body not shown in preview when email w/ att forwarded by iOS Mail client

SantoshKShetty:
- Initial commit


src/utils/normalize-mime-parts.ts
```diff
@@ -97,8 +101,31 @@ export function normalizeMimeParts(
 			if (disposition !== 'attachment') {
 				let bodyType =
 					type === 'text/html' ? 'html' : type === 'text/plain' && 'text';
-				if (bodyType && (!acc[bodyType] || disposition !== 'inline')) {
-					acc[bodyType] = content;
+
+				if (
```
yashrajpchavda: would that be a good idea to explain this flow in a comment? So that anyone who reads the code knows why this special case was written.

SantoshKShetty: Done

src/utils/normalize-mime-parts.ts
```diff
@@ -73,12 +73,17 @@ export function normalizeMimeParts(
...
 	{ origin, jwtToken }: { jwtToken?: string; origin?: string }
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = ({ ...attachment }, type: string) => {
 		attachment.messageId = attachment.messageId || message.id;
 		attachment.url = getAttachmentUrl(attachment, { origin, jwtToken });
-		if (attachment.contentId) {
-			attachment.contentId = normalizeCid(attachment.contentId);
-		}
+		attachment.contentId = attachment.contentId
```
itsmeprashant: Use grouping of operation, otherwise it will lead to ambiguity.

SantoshKShetty: eslint removes the grouping.

src/utils/normalize-mime-parts.ts
```diff
@@ -73,12 +73,17 @@ export function normalizeMimeParts(
 	message: { [key: string]: any },
 	{ origin, jwtToken }: { jwtToken?: string; origin?: string }
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = ({ ...attachment }, type: string) => {
```
pl12133: Do we need a second argument? I thought that `attachment` would have a property for `contentType` without any other arguments.

SantoshKShetty: Oh. That actually, I added just not to write same `normalizeType(attachment.contentType)` code. But, removed now.

PR #153, opened 2019-03-18
PREAPPS-2133: User Sessions | Send useragent data in header

madhavdhuppe:
Here we are using legacy user-agent mechanism to parse userAgent data.

link of legacy code:- https://github.com/Zimbra/zm-ajax/blob/develop/WebRoot/js/ajax/boot/AjxEnv.js


src/user-agent/index.ts
```diff
@@ -0,0 +1,402 @@
...
+	isIE6: false,
+	isIE6up: false,
+	isIE7: false,
+	isIE7up: false,
+	isIE8: false,
+	isIE8up: false,
+	isIE9: false,
+	isIE9up: false,
+	isIE10: false,
+	isIE12: false,
```
pl12133: Can we remove the browsers that we don't intend to support?

src/user-agent/index.ts
```diff
@@ -0,0 +1,402 @@
+export const userAgentData = {
```
pl12133: As property names the compiler cannot minify these variables, which could take a lot of space. Instead we can let them be scoped within the module:

```js
export let geckoDate = 0;
export let mozVersion = -1;
...etc...
```

The way the compiler could turn these into minified variable names.

src/batch-client/index.ts
```diff
@@ -141,8 +142,10 @@ export class ZimbraBatchClient {
 	private dataLoader: DataLoader<RequestOptions, RequestBody>;
 	private jwtToken?: string;
 	private notificationHandler?: NotificationHandler;
+	private serverVersion?: string;
 
 	constructor(options: ZimbraClientOptions = {}) {
+		parseUserAgent();
```
silentsakky: api-client should be not do any browser parsing and add data related to that in requests, client should parse it and pass details to api-client

src/user-agent/index.ts
silentsakky: can we explore using some external library for browser parsing

src/batch-client/index.ts
madhavdhuppe: will update

src/user-agent/index.ts
madhavdhuppe: Sure will look into it.

yashrajpchavda: @silentsakky, @madhavdhuppe had this idea, but we thought as we already have a proven solution, we should use it, as it would give us freedom to change it if necessary, but surely we can look at the alternatives and take a decision.

PR #158, opened 2019-04-04
local_folders_spikes: add query and resolver for local folders read

yashrajpchavda:


src/schema/schema.ts
```diff
@@ -78,6 +78,8 @@ export function createZimbraSchema(
 		notificationHandler: notifications && notifications.notificationHandler
 	});
 
+	const localStoreClient = options.localStoreClient;
```
pl12133: How will this look when we cut over to a real GraphQL backend? I think we may have to move this to the `@client` resolvers.

PR #161, opened 2019-04-04
PREAPPS-193 : Attachment Preview: Text files show spinner instead of preview

auminpatel:


src/batch-client/index.ts
```diff
@@ -1072,6 +1042,35 @@ export class ZimbraBatchClient {
 			return isError(response) ? [response] : [response.body];
 		});
 
+	private downlaod = ({ id, part, isSecure }: any) =>
```
silentsakky: typo: download

PR #163, opened 2019-04-08
PREAPPS-1055: Forward invite message as an event instead of as a message

khalidkadiri:
Forward invite message as an event instead of as a message

### DIFF
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -338,7 +338,6 @@ const ForwardMessageInput = new Entity({
 });
 
 export const ForwardAppointmentInviteInfo = new Entity({
-	id: 'id',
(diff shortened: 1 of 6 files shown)


src/normalize/entities.ts
```diff
@@ -331,6 +331,17 @@ Folder.addMapping({
...
 export { Folder };
 
+const ForwardMessageInput = new Entity({
+	e: ['emailAddresses', MailItemEmailAddress],
+	mp: ['mimeParts', MimePart],
+	su: 'subject'
+});
+
+export const ForwardAppointmentInviteInfo = new Entity({
+	id: 'id',
```
silentsakky: this mapping is not needed

khalidkadiri: Done

PR #173, opened 2019-04-29
PREAPPS-2261: User session appears multiple times in Settings/Sessions & Activity

madhavdhuppe:


src/schema/sessionHandler.ts
```diff
@@ -0,0 +1,38 @@
+import gql from 'graphql-tag';
```
silentsakky: don't use camelCase in file names

src/batch-client/types.ts
```diff
@@ -50,9 +50,23 @@ export type UserAgent = {
 	version: string;
 };
 
+export type writeSessionId = (sessionId: string) => void;
+
+export type readSessionId = () => string;
+
+export type Session = {
```
itsmeprashant: Where do we use this?

src/schema/session-handler.ts
```diff
@@ -0,0 +1,38 @@
...
+export class SessionHandler {
+	private cache: ZimbraInMemoryCache;
+	private sessionIdKey: string = 'sessionId';
+
+	constructor(options: ZimbraSessionOptions) {
+		this.cache = options.cache;
+	}
+
+	public readSessionId = () => {
+		const sessionIdFragment: any = this.cache.readFragment({
```
itsmeprashant: @silentsakky , Should we use `@client` directive or we are fine with ready and writing fragment directly?

src/batch-client/index.ts
```diff
@@ -1017,9 +1019,7 @@ export class ZimbraBatchClient {
 			const sessionId = get(response, 'header.context.session.id');
 			const notifications = get(response, 'header.context.notify.0');
 
-			if (sessionId) {
-				this.sessionId = sessionId;
-			}
+			this.checkAndUpdateSessionId(sessionId);
```
silentsakky: this function checks for sessionId === 1, in which response are we going to get 1 as session id?

src/batch-client/types.ts
silentsakky: we already have this as part of BaseRequestOptions#sessionId

src/batch-client/index.ts
```diff
@@ -1088,12 +1094,20 @@ export class ZimbraBatchClient {
...
 	 * These options are included on every request.
 	 */
-	private getAdditionalRequestOptions = () => ({
-		jwtToken: this.jwtToken,
-		sessionId: this.sessionId,
-		origin: this.origin,
-		userAgent: this.userAgent
-	});
+	private getAdditionalRequestOptions = () => {
+		if (this.sessionId === '1') {
```
silentsakky: instead of checking on default value, we can always read from cache for first time and define default value there
subsequent calls will read this.sessionId

silentsakky:
does this needs to be fixed in legacy zimbra-client as well? calls generating from there will still generate a new session with every call

madhavdhuppe:
> does this needs to be fixed in legacy zimbra-client as well? calls generating from there will still generate a new session with every call

yes, Same is happening with legacy i.e on reloading page it generates new sessionId

src/batch-client/index.ts
madhavdhuppe: updated

src/batch-client/types.ts
madhavdhuppe: updated

src/batch-client/index.ts
madhavdhuppe: updated

PR #176, opened 2019-05-06
PREAPPS-1831-2 Event notification shows for declined Calendar event

damini-vashishtha:


src/schema/schema.graphql
```diff
@@ -328,6 +328,14 @@ type Instance {
 	revision: Int # rev
 	status: InviteCompletionStatus
 }
+type Alarm {
```
silentsakky: please add mapping in normalize/entities.ts for keys which has shortforms

PR #177, opened 2019-05-06
PREAPPS-2376: Local Storage | Update mail list component to read data from the local store

yashrajpchavda:


yashrajpchavda:
Closed as the reference to the parent branch was not proper

PR #178, opened 2019-05-08
PREAPPS-2376: Local Storage | Update mail list component to read data from the local store

yashrajpchavda:


src/schema/schema.ts
```diff
@@ -117,8 +117,14 @@ export function createZimbraSchema(
 				getAvailableLocales: (_: any) => client.getAvailableLocales(),
 				getMailboxMetadata: (_: any, variables) =>
 					client.getMailboxMetadata(variables as GetMailboxMetadataOptions),
-				getMessage: (_, variables) =>
-					client.getMessage(variables as GetMessageOptions),
+				getMessage: (_, variables, context = {}) => {
```
yashrajpchavda: this code gets executed for rendering the mail detail panel, it's here as we have completed mail detail panel as well with hard coded data in this story

src/schema/schema.ts
```diff
@@ -117,8 +117,14 @@ export function createZimbraSchema(
...
+					const { local } = context;
```
pl12133: I'm still nervous about building on this pattern because this this schema will not exist when we cut over to a real graphql endpoint. If we keep going with this pattern it is tech debt waiting to happen.

We have to move all local operations into `@client` resolvers instead of regular resolvers.

yashrajpchavda: Sure Peter, will look at how we can achieve it without creating duplicate queries - one with @client and one without it. That's the only part which is preventing us to move to client resolvers I think.

yashrajpchavda: @pl12133 I will set up a call to discuss this in detail with you, we would finalise a strategy and handle this before we merge this code to develop (currently it would go to ZAPPD-38 branch)

PR #179, opened 2019-05-09
PREAPPS-2378: Local Storage | Update local folders component - Delete (move to Local Trash)

madhavdhuppe:


src/schema/schema.ts
```diff
@@ -190,8 +190,14 @@ export function createZimbraSchema(
 						}))
 			},
 			Mutation: {
-				action: (_, variables) => {
+				action: (_, variables, context) => {
```
yashrajpchavda: please pass a default value for context as `context = {}`, might fail when we try to access the `local` from context.

src/schema/schema.ts
```diff
@@ -190,8 +190,14 @@ export function createZimbraSchema(
...
 					const { type, ...rest } = variables;
+					const { local } = context;
+
+					if (local && type === 'FolderAction') {
```
yashrajpchavda: please remove the `FolderAction` check from here and add it in the local client if necessary as we would have only one handler which caters to all the actions in the local client.

src/schema/schema.ts
```diff
@@ -190,8 +190,14 @@ export function createZimbraSchema(
 						}))
 			},
 			Mutation: {
-				action: (_, variables) => {
+				action: (_, variables, context = {}) => {
 					const { type, ...rest } = variables;
+					const { local } = context;
```
pl12133: See https://github.com/Zimbra/zm-api-js-client/pull/178/files#r283481421

src/schema/schema.ts
```diff
@@ -196,8 +196,14 @@ export function createZimbraSchema(
...
+
+					if (local) {
```
itsmeprashant: return local ? localStoreClient.action(type, rest as ActionOptions) : client.action(type, rest as ActionOptions);

src/schema/schema.ts
```diff
@@ -196,8 +196,14 @@ export function createZimbraSchema(
 						}))
 			},
 			Mutation: {
-				action: (_, variables) => {
+				action: (_, variables, context = {}) => {
```
itsmeprashant: ( _, { type, ...rest }, context = {} )

PR #181, opened 2019-05-09
PREAPPS-2296: Account Recovery shows misleading error when password reset is disabled

pl12133:
Add `zimbraFeatureResetPasswordStatus` to AccountInfoAttrs.


src/schema/schema.graphql
```diff
@@ -1077,6 +1077,7 @@ type AccountInfoAttrs {
 	zimbraFeatureFiltersEnabled: Boolean
 	zimbraFeatureReadReceiptsEnabled: Boolean
 	zimbraFeatureManageZimlets: Boolean
+	zimbraFeatureResetPasswordStatus: String
```
silentsakky: this should be an Enum with possible values enabled, disabled, suspended

PR #183, opened 2019-05-15
PREAPPS-2494: Can't reply/forward in conversation view when Undo send is enable

SantoshKShetty:


src/normalize/entities.ts
```diff
@@ -220,7 +220,8 @@ export const SendMessageInfo = new Entity({
 });
 
 export const MessageInfo = new Entity({
-	...commonMailItemFields
+	...commonMailItemFields,
+	autoSendTime: 'autoSendTime'
```
silentsakky: you don't need entry in this file if key and value both are same

silentsakky: Fixed

PR #188, opened 2019-05-17
PREAPPS-2449: 2FA | Login Screen

SantoshKShetty:
- Initial commit


src/schema/schema.graphql
```diff
@@ -273,6 +273,7 @@ enum ZimletPresence {
 
 type AuthResponse {
 	authToken: [AuthToken]
+	twoFactorAuthRequired: TwoFactorAuthRequired
```
silentsakky: as per docs, type of this property should be Boolean, it's weird that current setup works

silentsakky: anyways api response confirms to implementation, no need to anything here

SantoshKShetty: Updated.

PR #190, opened 2019-05-21
PREAPPS-2246: Feature Enablement | SMIME

madhavdhuppe:
Added zimbraFeatureSMIMEEnabled in the account info.


silentsakky:
Not doing this implementation now

PR #195, opened 2019-05-24
PREAPPS-2500:Disable two factor auth.

anukritiJagatramka:


src/batch-client/index.ts
```diff
@@ -396,6 +396,12 @@ export class ZimbraBatchClient {
 			body: options
 		});
 
+	public disableTwoFactorAuth = () =>
+		this.jsonRequest({
+			name: 'DisableTwoFactorAuth',
+			namespace: Namespace.Account
+		}).then(() => true);
```
SantoshKShetty: Is returning true required ?

anukritiJagatramka: Removed explicit returning true.

PR #196, opened 2019-05-27
PREAPPS-2384: Local Storage | Implement drag and drop of an email item from online folder to a local folder

yashrajpchavda:


src/apollo/offline-queue-link/index.ts
```diff
@@ -113,11 +113,12 @@ export class OfflineQueueLink extends ApolloLink {
...
 			skipQueue,
 			cancelQueues,
-			offlineQueueName
+			offlineQueueName,
+			local
 		} = operation.getContext();
 
 		const isForwarding =
-			this.isOpen || skipQueue || hasSensitiveVariables(operation);
+			this.isOpen || local || skipQueue || hasSensitiveVariables(operation);
```
yashrajpchavda: this is added so that offline queue is bypassed for local folder operations

src/schema/schema.graphql
```diff
@@ -2684,6 +2703,7 @@ type Mutation {
 	resetPassword(password: String!): Boolean
 	revokeRights(input: RevokeRightsInput!): RevokeRightsResponse
 	saveDraft(message: SendMessageInput!): SaveDraftResponse
+	saveMessagesToLocal(messagesData: SaveMessagesToLocalInput!): SaveEmailsToLocalResponse
```
silentsakky: isn't this same as movemail operation? do we need extra mutation just for local folders?

src/schema/schema.graphql
```diff
@@ -2344,13 +2345,17 @@ input DismissInput {
...
 input AddMsgInput {
 	folderId: ID!
-	content: StringContentInput
+	absFolderPath: String
+	content: String
+	meta: String
+}
+
+input SaveMessageDataInput {
+	id: ID!,
```
silentsakky: commas are not needed

src/schema/schema.graphql
yashrajpchavda: removed

src/schema/schema.graphql
yashrajpchavda: fixed

PR #200, opened 2019-05-30
PREAPPS-2567: Zimbra X cannot read signed messages originating from Outlook users

SantoshKShetty:


src/batch-client/index.ts
```diff
@@ -127,9 +127,16 @@ function normalizeMessage(
 	{ origin, jwtToken }: { jwtToken?: string; origin?: string }
 ) {
 	const normalizedMessage = normalize(MessageInfo)(message);
-	normalizedMessage.attributes =
-		normalizedMessage.attributes &&
-		mapValuesDeep(normalizedMessage.attributes, coerceStringToBoolean);
+	const { isSigned, isEncrypted } = normalizedMessage; // GetMsgResponse itself returns `isSigned` and `isEncrypted`. Let's use them as defaults.
```
silentsakky: most probably we are getting this information because we have server side smime enabled
we will be disabling it, so this fix will not be valid in that case

SantoshKShetty: Awe.. didn't know that. Will have to figure out if there are any other solution for this.

SantoshKShetty: **UPDATE:**
I disabled smime zimlet which was enabled on an a/c where this issue was happening. And GetMsgResponse started returning correct data. So, this code change is not required at all.

PR #201, opened 2019-05-31
PREAPPS-2374: Local Storage | Forward an email present in local folder - without attachments

sachinpatel88:


src/schema/schema.graphql
```diff
@@ -458,6 +458,7 @@ type MessageInfo implements MailItem {
 	replyType: String #rt
 	attributes: MessageAttributes
 	autoSendTime: Float
+	local: Boolean
```
silentsakky: why do we need to pass this to graphql query/mutation, i thought this is only passed as context argument

yashrajpchavda: We have added an attribute on the MailItem which denotes whether that email is a local email or remote - ideally this flag would have same value as its parent folder's local attribute value. This is useful in performing tasks at the mail item level to check whether that email is local email or remote, such as forwarding, replying, etc.

PR #203, opened 2019-06-04
PREAPPS-2512 Expose method to update user agent string from clients

silentsakky:


src/batch-client/index.ts
```diff
@@ -1017,6 +1017,10 @@ export class ZimbraBatchClient {
 			mapValuesDeep(get(res, `${accountType}.0`), coerceStringToBoolean)
 		);
 
+	public updateUserAgent = (userAgent: Object) => {
```
pl12133: Can we name this `setUserAgent`?

PR #205, opened 2019-06-10
PREAPPS-2454: 2FA | One time codes

panditsavitags:


src/schema/schema.graphql
```diff
@@ -290,6 +290,10 @@ type ScratchCodeType {
 	_content: String
 }
 
+type ScratchCodes {
+	scratchCodes: ScratchCode
```
silentsakky: shouldn't this be an array? [ScratchCode]

src/batch-client/index.ts
```diff
@@ -614,6 +626,18 @@ export class ZimbraBatchClient {
 			body: denormalize(GetRightsRequest)(options)
 		}).then(normalize(AccountRights));
 
+	public getScratchCodes = (username: String) =>
+		this.jsonRequest({
+			name: 'GetScratchCodes',
+			namespace: Namespace.Account,
+			body: {
```
SantoshKShetty: Is this payload required ? API doesn't say so...

Also, I doubt `namespace` is required here.

src/batch-client/index.ts
```diff
@@ -489,6 +489,18 @@ export class ZimbraBatchClient {
 			}
 		}).then(res => normalize(FreeBusy)(res.usr));
 
+	public generateScratchCodes = (username: String) =>
+		this.jsonRequest({
+			name: 'GenerateScratchCodes',
+			namespace: Namespace.Account,
+			body: {
```
SantoshKShetty: Same as https://github.com/Zimbra/zm-api-js-client/pull/205/files#r292307415

src/batch-client/index.ts
panditsavitags: yes, I tried this api without this payload, it won't work so i checked the api call that is being made by legacy which includes this payload. So without the account name this api does not work as expected

panditsavitags: the default namespace is mail, and this call is related to account so i added the namespace

panditsavitags: https://files.zimbra.com/docs/soap_api/8.8.8/api-reference/index.html , the namesapace mentioned is zimbraAccount

src/batch-client/index.ts
silentsakky: username is mandatory parameter as per soap api

src/schema/schema.graphql
pl12133: It looks like the API doesn't return an Array, https://files.zimbra.com/docs/soap_api/8.8.8/api-reference/zimbraAccount/GenerateScratchCodes.html

silentsakky: Yeah, weird implementation of apis :)

PR #208, opened 2019-06-11
PREAPPS-2476 CSRF token support for Zimbra X (Do not merge the code to develop)

itsmeprashant:


src/request/index.ts
```diff
@@ -218,6 +220,11 @@ export function jsonRequest(
 		[soapRequestName]: soapCommandBody(options)
 	};
 
+	if (csrfToken) {
```
silentsakky: maybe we should move this to line 217

src/utils/csrf-token.ts
```diff
@@ -0,0 +1,7 @@
+export function getCSRFToken(): string | null {
```
silentsakky: should we move this write/read operation to zm-x-web? like how userAgent is managed

itsmeprashant: it is better if we keep token related logic here, cz it will only be used in api client

src/request/index.ts
```diff
@@ -243,3 +244,21 @@ export function jsonRequest(
 			};
 		});
 }
+
+export function zmFetch(url: string, request: any = {}): Promise<Response> {
```
pl12133: Can we add a comment that the reason this is here is for CSRF protection?

src/utils/csrf-token.ts
```diff
@@ -0,0 +1,7 @@
...
+	return localStorage.getItem('CSRFToken');
```
pl12133: This is a little different from legacy since we don't use an HTTPOnly cookie, but I think it should still be OK.

The only attack I can think of would be a Zimlet stealing the CSRF token in order to forge requests, but at that point the Zimlet already has the power to make API requests with the users real auth token so it doesn't matter.

PR #214, opened 2019-06-19
PREAPPS-2616: Intermittent script error: Uncaught (in promise) Error: GraphQL error: Syntax Error: Invalid number, unexpected digit after 0: "9" for some operations

madhavdhuppe:
used genarateFragmentName function to avoid dash and colon in the fragment name

### DIFF
--- a/src/schema/notifications.ts
+++ b/src/schema/notifications.ts
@@ -114,8 +114,7 @@ export class ZimbraNotifications {
 					`
 					});
 				} catch (exception) {
-					console.warn(exception);
-					return;
+					console.error(exception);
 				}
 				const folderName = (folder && folder.name) || defaultFolderName;
 				const group =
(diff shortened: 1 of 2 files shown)


src/schema/notifications.ts
```diff
@@ -103,14 +103,19 @@ export class ZimbraNotifications {
...
-					fragment: gql`
-					fragment folderName${item.folderId} on Folder {
-						name
-					}
+				let folder: any;
+				try {
+					folder = this.cache.readFragment({
+						id: `Folder:${item.folderId}`,
+						fragment: gql`
+						fragment ${generateFragmentName('folderName', item.folderId)} on Folder {
```
silentsakky: where are we normalizing name when writing to cache? or the cache will have : in folder id?

madhavdhuppe: Yes folder id is coming with ":"

PR #218, opened 2019-07-03
PREAPPS-2705: Local Storage | Allow local emails to be moved to remote folders

SantoshKShetty:


src/batch-client/index.ts
```diff
@@ -214,16 +214,26 @@ export class ZimbraBatchClient {
 			}
 		}).then(res => get(res, `${accountType}.0.id`));
 
-	public addMessage = ({ folderId, content }: AddMsgInput) =>
-		this.jsonRequest({
+	public addMessage = (options: AddMsgInput) => {
+		const { folderId, content, meta } = get(options, 'message');
+		const { flags, tags, tagNames, date } = JSON.parse(meta);
```
yashrajpchavda: parsing should ideally be part of try, catch, just to be safe

SantoshKShetty: Done

PR #220, opened 2019-07-05
PREAPPS-1267: Actions for GAL as a "System Folder" in contacts

sachinpatel88:


src/schema/notifications.ts
```diff
@@ -131,17 +131,22 @@ export class ZimbraNotifications {
...
 				const { sortBy }: any = getVariablesFromDataId(id) || {};
 				if (!searchResponse[query] && id) {
-					searchResponse[query] = this.cache.readFragment({
-						id: id,
-						fragment: gql`
-							fragment ${generateFragmentName('searchResults')} on SearchResponse {
-								contacts {
-									id
-									fileAsStr
+					try {
```
yashrajpchavda: can we add a comment explaining why we wrap read fragments in try catch? It can serve as learning to the ones who read comments ;)

PR #228, opened 2019-07-23
PREAPPS-2289:  User uploaded profile picture/avatar flickers during page load

labderahul1:


src/utils/normalize-mime-parts.ts
```diff
@@ -64,7 +64,7 @@ export function getProfileImageUrl(
 ) {
 	return `${origin}/service/home/~/?max_width=100&max_height=100&auth=${
 		jwtToken ? 'jwt' : 'co'
-	}&id=${encodeURIComponent(profileImageId)}&t=${Date.now()}${
+	}&id=${encodeURIComponent(profileImageId)}${
```
silentsakky: this means we have removed cache busting part, is this going to work even when images gets updated?

labderahul1: Yes. For new image generate new/unique profileImageId.

itsmeprashant: @silentsakky do you want to keep this change?

itsmeprashant: does this gernerate new id for updated images?

silentsakky: if we are having different profileImageId when updating image then it is fine to remove cache busting param

PR #238, opened 2019-07-30
PREAPPS-2941: Tags | Display tags in side bar

avadhutprabhudesai:


src/batch-client/index.ts
```diff
@@ -707,6 +708,20 @@ export class ZimbraBatchClient {
 			namespace: Namespace.Account
 		});
 
+	public getTag = () =>
+		this.jsonRequest({
+			name: 'GetTag',
+			namespace: Namespace.Mail
+		}).then(({ tag }) => {
```
yashrajpchavda: can we write this like,

`}).then(({ tag }) => tag.map(normalize(Tag)) || []);`

src/schema/schema.ts
```diff
@@ -177,7 +177,8 @@ export function createZimbraSchema(
 					client.shareInfo(variables as ShareInfoOptions),
 				taskFolders: client.taskFolders,
 				getWhiteBlackList: client.getWhiteBlackList,
-				getAppSpecificPasswords: client.getAppSpecificPasswords
+				getAppSpecificPasswords: client.getAppSpecificPasswords,
+				getTag: () => client.getTag()
```
itsmeprashant: getTag: client.getTag

src/batch-client/index.ts
avadhutprabhudesai: Is it good idea to modify function arguments directly?

yashrajpchavda: Umm, not sure if I understand you correctly, can you please elaborate?

avadhutprabhudesai: Got it. Fixed.

src/schema/schema.ts
avadhutprabhudesai: Fixed.

src/schema/schema.graphql
```diff
@@ -1394,6 +1394,13 @@ type ContactListMember {
 	value: ID!
 }
 
+type Tag {
+	id: String
```
SantoshKShetty: Shouldn't this be `ID` ?

avadhutprabhudesai: Fixed.

PR #239, opened 2019-08-01
PREAPPS-2704: Local Storage | Implement permanent delete (delete an email from local trash)

sachinpatel88:


src/schema/schema.ts
```diff
@@ -321,7 +321,13 @@ export function createZimbraSchema(
 
 				changeFolderColor: (_, variables) =>
 					client.changeFolderColor(variables as FolderActionChangeColorInput),
-				folderAction: (_, { action }) => client.folderAction(action),
+				folderAction: (_, { action, context }) => {
```
itsmeprashant: `folderAction: (_, { action, context }) => context && context.local ? localStoreClient.action() : client.folderAction(action),`

src/schema/schema.ts
```diff
@@ -321,7 +321,10 @@ export function createZimbraSchema(
 
 				changeFolderColor: (_, variables) =>
 					client.changeFolderColor(variables as FolderActionChangeColorInput),
-				folderAction: (_, { action }) => client.folderAction(action),
+				folderAction: (_, { action, context }) =>
```
yashrajpchavda: if I am not mistaken, actually `context` comes as a second parameter in the resolver function, previously, have we done anything different for folder action mutation to make `context` part of `variables` itself?

yashrajpchavda: Sorry, I missed it earlier, but for other folder related actions (for local as well as remote), we are using `action` mutation from this file, does the remote call to permanently delete a folder, use this method?

sachinpatel88: Yeah It's not being used. I think I need to delete this branch. :D

sachinpatel88:
Delete as folderAction mutation changes not required.

PR #241, opened 2019-08-01
PREAPPS-2821: Break account info query into multiple queries

lavegupta:


src/batch-client/index.ts
```diff
@@ -175,26 +175,20 @@ export class ZimbraBatchClient {
...
-					license: {
-						status: res.license.status,
-						attr: mapValuesDeep(res.license.attr, coerceStringToBoolean)
-					}
-				})
-			};
-		});
+			namespace: Namespace.Account,
+			body: {
+				sections: 'mbox,attrs,zimlets'
```
itsmeprashant: FYI - This is to define the kind of data expected from response of GetInfoRequest.

src/batch-client/index.ts
```diff
@@ -584,6 +578,11 @@ export class ZimbraBatchClient {
 			return c;
 		});
 
+	public getDataSources = () =>
```
itsmeprashant: FYI - To get IMAP and POP account data.

src/batch-client/index.ts
```diff
@@ -175,26 +175,20 @@ export class ZimbraBatchClient {
...
-					}
-				})
-			};
-		});
+			namespace: Namespace.Account,
+			body: {
+				sections: 'mbox,attrs,zimlets'
+			}
+		}).then(res => ({
+			...res,
```
itsmeprashant: If we are changing return data here then we should update `type AccountInfo`, and entity if needed.

lavegupta: Not needed.

PR #246, opened 2019-08-14
PREAPPS-3044: Reduce refetch calls and implement server side notification for drafts

SantoshKShetty:


index.ts
```diff
@@ -11,3 +11,17 @@ export { SyncOfflineOperations } from './src/apollo/sync-offline-operations';
 export { batchJsonRequest, jsonRequest } from './src/request';
 import * as types from './types';
 export { types };
+export { normalize } from './src/normalize';
+import {
```
yashrajpchavda: can do direct export, like done above.

SantoshKShetty: Actually, I wanted only few things without introducing conflicts, so used namespaces

PR #248, opened 2019-08-19
PREAPPS-1180: "Run" action for existing filters 

craigdanj:


package-lock.json
```diff
@@ -2036,7 +2036,7 @@
     },
```
madhavdhuppe: I think we are not updating package-lock.json files if we haven’t updated/added npm packages.

craigdanj: you are right. This was done by mistake. Will undo it.

craigdanj: Done

src/batch-client/index.ts
```diff
@@ -247,6 +248,28 @@ export class ZimbraBatchClient {
...
+			body: {
+				filterRules: {
+					filterRule: filterRules
+				},
+				m: {
+					ids
+				}
+			}
+		}).then(res => {
+			let ids = get(res, 'm[0].ids');
```
silentsakky:
```
const ids = get(res, 'm[0].ids');

			if (ids) {
				return ids.split(',');
			}

			return [];
```

craigdanj: done

src/batch-client/index.ts
```diff
@@ -247,6 +248,27 @@ export class ZimbraBatchClient {
...
+					filterRule: filterRules
+				},
+				m: {
+					ids
+				}
+			}
+		}).then(res => {
+			const ids = get(res, 'm[0].ids');
+
+			if (ids) {
```
avadhutprabhudesai: return ids ? ids.split(',') : [];

craigdanj: done

src/batch-client/index.ts
```diff
@@ -241,6 +242,22 @@ export class ZimbraBatchClient {
...
+			name: 'ApplyFilterRules',
+			body: {
+				filterRules: {
+					filterRule: filterRules
+				},
+				m: {
+					ids
+				}
+			}
+		}).then(res => {
```
itsmeprashant: then(res => get(res, 'm[0].ids', []).split(','));

itsmeprashant: never mind it will not work, I thought of list when I commented.

PR #251, opened 2019-08-23
PREAPPS-1072: Deleting/moving multiple messages operation extremely slow

madhavdhuppe:


src/batch-client/index.ts
```diff
@@ -1185,8 +1185,10 @@ export class ZimbraBatchClient {
 
 			this.checkAndUpdateSessionId(sessionId);
 
-			if (notifications && this.notificationHandler) {
-				this.notificationHandler(notifications);
+			if (notifications) {
+				setTimeout(() => {
```
BhavinBathani: can you please add an inline comment describing the reason ?

madhavdhuppe: added

PR #252, opened 2019-08-23
PREAPPS-2589: Email folder sharing

SantoshKShetty:


src/batch-client/index.ts
```diff
@@ -595,7 +595,56 @@ export class ZimbraBatchClient {
...
+		}).then(res => {
+			const foldersResponse = normalize(Folder)(res);
+			const folders = get(foldersResponse, 'folders.0', {});
+
+			if (folders.linkedFolders) {
+				/**
+				 * This function is required because the API returns Subfolder data for shared folder
+				 * with Actual folder path (not mounted folder path). This could lead to 404 "NO SUCH FOLDER EXISTS ERROR".
+				 */
+				const updateAbsoluteFolderPath = (
```
BhavinBathani: Don't you think this function would be declared outside as a member of this class it self?

SantoshKShetty: No. That's very specific to share EMAIL (Could be for Contact) folders. So kept it as local function.

yashrajpchavda: The only problem that I see is we declare it every time when the callback for getFolder executes, and then it gets garbage collected, again declared and collected every time when the query is done (which is many times I think). I'd propose to move it :)

SantoshKShetty: Done. Thanks

PR #257, opened 2019-08-28
PREAPPS-3202 Don't add csrftoken when sending auth request

silentsakky:


src/batch-client/index.ts
```diff
@@ -1178,7 +1179,10 @@ export class ZimbraBatchClient {
 	private batchDataHandler = (requests: Array<RequestOptions>) =>
 		batchJsonRequest({
 			requests,
-			...this.getAdditionalRequestOptions()
+			// check if login request then don't add csrfToken
+			...this.getAdditionalRequestOptions(
```
silentsakky: this is not required now, as we are doing singleRequest for auth

PR #258, opened 2019-08-28
PREAPPS-3062: Sharing contact folders

khalidkadiri:


src/batch-client/index.ts
```diff
@@ -628,9 +628,22 @@ export class ZimbraBatchClient {
...
+					) {
+						const {
+							absFolderPath,
+							oname,
+							folders,
+							ownerZimbraId,
+							sharedItemId
+						} = folder;
+
+						if (folder.view === FolderView.Contact) {
```
SantoshKShetty: Let's add comment on why we're doing this for 1st level of linkedFolders only.

khalidkadiri: Added comment

PR #260, opened 2019-09-03
ZXL-107 Front End Performance Improvements

itsmeprashant:


src/schema/notifications.ts
```diff
@@ -96,8 +98,78 @@ export class ZimbraNotifications {
...
+		const ITERATIONS = Math.ceil(LENGTH / BATCH_SIZE);
+
+		let i: number;
+		for (i = 0; i < ITERATIONS; i++) {
+			const start = i * BATCH_SIZE;
+			let end = start + BATCH_SIZE;
+
+			if (end > LENGTH) {
+				end = LENGTH;
+			}
```
pl12133: Can combine the above 5 lines:

```js
const end = Math.min(start + BATCH_SIZE, LENGTH);
```

src/schema/notifications.ts
```diff
@@ -96,8 +98,78 @@ export class ZimbraNotifications {
...
+						this.broadcastCacheUpdates();
+					}
+				})(i, ITERATIONS, batch),
+				TIMEOUT
+			);
+		}
+	}
+
+	private broadcastCacheUpdates = () => {
+		this.getApolloClient().queryManager.broadcastQueries();
```
pl12133: Does calling `writeFragment` already do this?

itsmeprashant: Here I think problem is with cache, we perform `writeFragment` on cache object and which is not broadcasting updated data instantaneously,  https://www.apollographql.com/docs/react/advanced/caching/#direct-cache-access

PR #277, opened 2019-10-17
Zx 1070 Create a means to add a Skype deep link for a meeting

lgregoire:


src/batch-client/index.ts
```diff
@@ -1056,6 +1056,16 @@ export class ZimbraBatchClient {
 		});
 	};
 
+	public modifyProps = (props: any) =>
```
DeanoDee: Maybe it's just me, but the name of this method is confusing to me. The reason is props in the react preact world should usually be treated as immutable, sure you can calculate other values but the best practice, this out of practice dev, followed was don't tough them. So is this props from react? I'm not sure what this is doing, could a bit more verbosity improve context (parsePropsForJava - maybe)?

ronstra-synacor: They're actually _Zimlet_ properties. I agree the name is a bit generic, like "data" or "value", but I assume that was because they're just called plain old "Properties" in the SOAP API this abstracts.

PR #278, opened 2019-10-18
PREAPPS-3593: Folder count issue fixed for shared folder.

anukritiJagatramka:


src/schema/notifications.ts
```diff
@@ -302,8 +322,12 @@ export class ZimbraNotifications {
 		if (items) {
 			items.forEach((i: any) => {
 				const item = normalizeFolder(i);
+				let itemId: any;
```
SantoshKShetty: There was another PR - https://github.com/Zimbra/zm-api-js-client/pull/270. Aren't we proceeding with that ?

src/schema/notifications.ts
```diff
@@ -145,6 +145,26 @@ export class ZimbraNotifications {
...
 
+	// Find the actual folder of the shared folder
+	private findSharedItemId = (item: any) => {
+		let sharedFolderId: any;
+		const cachedData = get(this.cache, 'data.data');
+		const allFolders = Object.keys(cachedData).filter(f =>
+			f.includes('Folder:')
+		);
+		const idSplit = item.id.split(':');
+		for (let folderId in allFolders) {
```
avadhutprabhudesai: can folderId be const?

src/schema/notifications.ts
```diff
@@ -145,6 +145,26 @@ export class ZimbraNotifications {
...
+			if (
+				cachedData[allFolders[folderId]].ownerZimbraId === idSplit[0] &&
+				cachedData[allFolders[folderId]].sharedItemId === idSplit[1]
+			) {
+				sharedFolderId = cachedData[allFolders[folderId]].id;
```
avadhutprabhudesai: Might want to explain break condition in comment.

src/schema/notifications.ts
```diff
@@ -145,6 +145,26 @@ export class ZimbraNotifications {
...
+	private findSharedItemId = (item: any) => {
+		let sharedFolderId: any;
+		const cachedData = get(this.cache, 'data.data');
+		const allFolders = Object.keys(cachedData).filter(f =>
+			f.includes('Folder:')
+		);
+		const idSplit = item.id.split(':');
+		for (let folderId in allFolders) {
+			if (
+				cachedData[allFolders[folderId]].ownerZimbraId === idSplit[0] &&
```
avadhutprabhudesai: Store cachedData[allFolders[folderId]] into a variable and use it at multiple places.

src/schema/notifications.ts
```diff
@@ -145,6 +145,26 @@ export class ZimbraNotifications {
 		this.getApolloClient().queryManager.broadcastQueries();
 	};
 
+	// Find the actual folder of the shared folder
+	private findSharedItemId = (item: any) => {
```
avadhutprabhudesai: Can we destructure id here?

src/schema/notifications.ts
```diff
@@ -302,8 +322,12 @@ export class ZimbraNotifications {
 		if (items) {
 			items.forEach((i: any) => {
 				const item = normalizeFolder(i);
+				let itemId = item.id;
+				if (i.id.includes(':')) {
+					itemId = this.findSharedItemId(i);
```
avadhutprabhudesai: Can we just pass id to this function? Do we need any other attributes from item?

src/schema/notifications.ts
```diff
@@ -302,8 +322,12 @@ export class ZimbraNotifications {
 		if (items) {
 			items.forEach((i: any) => {
 				const item = normalizeFolder(i);
+				let itemId = item.id;
```
avadhutprabhudesai: const itemId = i.id.includes(':') ? this.findSharedItemId(i.id) : item.id; ?

src/schema/notifications.ts
itsmeprashant: `allFolders[folderId][ownerZimbraId]` holds same data as `cachedData[allFolders[folderId]].ownerZimbraId `

src/schema/notifications.ts
itsmeprashant: sharedFolderId = allFolders[folderId][id]

src/schema/notifications.ts
```diff
@@ -302,8 +322,12 @@ export class ZimbraNotifications {
...
+				if (i.id.includes(':')) {
```
itsmeprashant: `get(i, 'id', '').includes(':')`

src/schema/notifications.ts
anukritiJagatramka: The approch was changed, incorporated the same.

src/schema/notifications.ts
anukritiJagatramka: Done.

src/schema/notifications.ts
anukritiJagatramka: @itsmeprashant `allFolders `is an array of keys of type `Folder:`, for example - `['Folder:2', 'Folder:1999']`. It is different from cached data.
@avadhutprabhudesai Added comment.

src/schema/notifications.ts
anukritiJagatramka: @itsmeprashant As mentioned above, allFolders is an array of string of type `Folder:`.
@avadhutprabhudesai Stored the data in variable and used it everywhere.

src/schema/notifications.ts
anukritiJagatramka: Tweaked logic to only send id to `findSharedItemId`, hence destructuring will not be required.

src/schema/notifications.ts
anukritiJagatramka: Done.

src/schema/notifications.ts
anukritiJagatramka: Done.

src/schema/notifications.ts
anukritiJagatramka: Done.

src/schema/notifications.ts
```diff
@@ -145,6 +145,28 @@ export class ZimbraNotifications {
...
+		const idSplit = item.split(':');
+		for (const folderId in allFolders) {
+			const folder: any = cachedData[allFolders[folderId]];
+			//Find the folder where ownerZimbraId:sharedItemId equals to id
+			if (
+				folder.ownerZimbraId === idSplit[0] &&
+				folder.sharedItemId === idSplit[1]
+			) {
+				sharedFolderId = folder.id;
+				break;
```
pl12133: Instead of doing an assignment and a `break` here, you can instead `return folder.id;` and won't need to break out of the loop because the function ends.

src/schema/notifications.ts
```diff
@@ -302,8 +324,11 @@ export class ZimbraNotifications {
 		if (items) {
 			items.forEach((i: any) => {
 				const item = normalizeFolder(i);
+				const itemId = get(i, 'id', '').includes(':')
```
pl12133: We can access `item.id` here safely instead of using `get`, and it is confusing to use `i` instead of `item`.

anukritiJagatramka: Done.

src/schema/notifications.ts
anukritiJagatramka: Done.

PR #280, opened 2019-10-23
graphql-preactx Upgrade graphql packages

silentsakky:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@zimbra/api-client",
   "amdName": "zmApiJsClient",
-  "version": "25.7.0",
+  "version": "25.7.0-beta.1",
```
silentsakky: do we need this?

PR #298, opened 2019-12-18
PREAPPS-2077: Forward event from Calendar context menu

craigdanj:


src/normalize/entities.ts
```diff
@@ -349,6 +350,11 @@ const ForwardMessageInput = new Entity({
 	su: 'subject'
 });
 
+export const ForwardAppointmentInfo = new Entity({
+	m: ['message', ForwardMessageInput],
+	exceptId: ['exceptId', InstanceDate]
```
SantoshKShetty: `exceptId` is not required to be included here.

craigdanj: not sure I understood. We have to pass exceptId when we want to forward an instance right?

craigdanj: As discussed with you.. this is needed hence we need to keep it.

SantoshKShetty: Sorry. Missed above `InstanceDate`. Thanks @craigdanj

PR #304, opened 2020-01-07
PREAPPS-1860: Show notifications when new messages arrive

aayushvijayvergiya:
New mail(only with flag 'unread') received in notifications object is written to local cache using `@client` directive. Client query listening to the same is updated with new mail.

### DIFF
--- a/src/schema/notifications.ts
+++ b/src/schema/notifications.ts
@@ -20,6 +20,16 @@ const normalizeFolder = normalize(FolderEntity);
 const normalizeMessage = normalize(MessageInfo);
 const normalizeContact = normalize(Contact);
 const normalizeTag = normalize(Tag);
+const writeNewMailQuery = gql`
+	query getNewMail {
+		getNewMail @client {
+			id
+			subject
+			flags
+			folderId
+		}
+	}
+`;
 
 function itemsForKey(notification: any, key: string) {
 	const modifiedItems = get(notification, `modified.${key}`, []);
@@ -409,6 +419,7 @@ export class ZimbraNotifications {
 					}
 				});
 			});
+			items.length && this.writeToCacheForNewMail(items);
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


src/schema/schema.graphql
```diff
@@ -2731,6 +2742,8 @@ type Query {
 	taskFolders: [Folder]
 
 	getTag: [Tag]
+
+	getNewMail: NewMail
```
itsmeprashant: Shouldn't it be an array, what if more then one emails arrive at the same time

aayushvijayvergiya: Yes, that should have been an array, changed. Also changed related logic.

src/schema/notifications.ts
```diff
@@ -431,4 +442,28 @@ export class ZimbraNotifications {
...
+					flags.indexOf('u') > -1 &&
+					itemsToWrite.push({
+						id: item.id,
+						subject: item.subject,
+						flags: item.flags || item.flag,
+						folderId: item.folderId,
+						__typename: 'NewMail'
+					});
+			});
+		this.cache.writeQuery({
```
silentsakky: we should check if itemsToWrite contains any item or not

aayushvijayvergiya: Added a check at https://github.com/Zimbra/zm-api-js-client/pull/304/files#diff-50cec4d28ea90d367f48b5a273ce67aaR422 so that the function itself won't be called if no items are present.

src/schema/notifications.ts
```diff
@@ -431,4 +442,35 @@ export class ZimbraNotifications {
...
+
+	private writeToCacheForNewMail = (items: any) => {
+		let itemsToWrite: any = [];
+		let data: any;
+		try {
+			data = this.cache.readQuery({ query: writeNewMailQuery });
+		} catch (exception) {
+			console.error(exception);
+		}
+		data && (itemsToWrite = data.getNewMail);
```
itsmeprashant:
```
let itemsToWrite: any = [];
try {
    itemsToWrite = this.cache.readQuery({ query: writeNewMailQuery }).getNewMail;
} catch (exception) {
    itemsToWrite = [];
    console.error(exception);
}
```

PR #307, opened 2020-01-16
PREAPPS-3706: Provide an option to switch to Legacy UI from Zimbra X UI

madhavdhuppe:
Added zimbraPrefClientType attribute into the preferences and preferencesInput.

### DIFF
--- a/src/schema/schema.graphql
+++ b/src/schema/schema.graphql
@@ -289,6 +289,12 @@ enum ResetPasswordStatus {
 	suspended
 }
 
+enum PrefClientType {
+	advanced,
+	zimbrax,
+	standard
+}
+
 type CsrfToken {
 	_content: String
 }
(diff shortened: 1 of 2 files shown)


src/schema/schema.graphql
```diff
@@ -1289,6 +1289,7 @@ type Preferences {
 	zimbraPrefWebClientOfflineBrowserKey: String
 	zimbraPrefTimeZoneId: String
 	zimbraPrefLocale: String
+	zimbraPrefClientType: String
```
silentsakky: should be an enum

madhavdhuppe: updated

src/schema/schema.graphql
```diff
@@ -289,6 +289,12 @@ enum ResetPasswordStatus {
 	suspended
 }
 
+enum PrefClientType {
+	advanced,
```
silentsakky: this is confusing, we have two options only, right?

madhavdhuppe: we are getting three options from API response.

PR #309, opened 2020-01-29
PREAPPS-3064: Inessential requests get fired on navigate to Mail Tab

SantoshKShetty:


src/schema/notifications.ts
```diff
@@ -205,187 +215,216 @@ export class ZimbraNotifications {
 	};
 
 	private processContactNotifications = (items: any) => {
-		if (items) {
```
yashrajpchavda: Curious to know why have we done this change.

SantoshKShetty: This is already handled in `batchProcessItems`. Hence did that.

yashrajpchavda: aah okay

PR #312, opened 2020-02-05
PREAPPS-4167: Inessential requests get fired on navigate to Calendar Tab

SantoshKShetty:


src/batch-client/index.ts
```diff
@@ -882,6 +882,27 @@ export class ZimbraBatchClient {
 			namespace: Namespace.Mail
 		}).then(({ tag = [] }) => tag.map(normalize(Tag)));
 
+	public getTasks = (options: SearchOptions) =>
```
silentsakky: didn't we already have query to get task data?
and also we should map query with soap api endpoint

SantoshKShetty: Hmm.. It seems so.. Need to find out why new query was used.

SantoshKShetty: @silentsakky there's a TaskQuery but its different (being used to get task data for given task). This is a search task request.

itsmeprashant: We are doing same thing here: 
https://github.com/Zimbra/zm-api-js-client/blob/develop/src/schema/schema.ts#L220

Then why this chenage is needed?

src/schema/schema.ts
```diff
@@ -199,41 +197,6 @@ export function createZimbraSchema(
 					return obj.conversationId ? 'MessageInfo' : 'Conversation';
 				}
 			},
-			Folder: {
```
itsmeprashant: Why did we remove this?

SantoshKShetty: Found no usage of it.

yashrajpchavda: @SantoshKShetty should we remove the respective `folder -> appointments` mapping from schema.graphql too?

SantoshKShetty: Nice catch. Fixed.

SantoshKShetty: For now, I have removed only `folder -> tasks` mapping from schema.graphql as we're still using CalendarsAndAppointmentsQuery in new-event-from-email zimlet.

Once we refactor that zimlet, we'll update schema.graphql and remove `folder -> appointments`

PR #314, opened 2020-02-11
PREAPPS-4164: System sends multiple requests for snooze all or dismiss all

sachinpatel88:


src/schema/notifications.ts
```diff
@@ -485,7 +485,6 @@ export class ZimbraNotifications {
 			itemsToWrite = data.getNewMail;
 		} catch (exception) {
 			itemsToWrite = [];
-			console.error(exception);
```
sachinpatel88: It's known behaviour so not required to console it. (Change suggested in PREAPPS-1860)
cc: @aayushvijayvergiya @itsmeprashant

silentsakky: any reason for removing this?

src/schema/schema.graphql
```diff
@@ -2884,8 +2884,8 @@ type Mutation {
 	sendInviteReply(inviteReply: InviteReplyInput!): InviteReplyResponse
 	sendShareNotification(shareNotification: ShareNotificationInput!): Boolean
 	setMailboxMetadata(section: String, attrs: MailboxMetadataSectionAttrsInput!): Boolean
-	snoozeCalendarItem(appointment: SnoozeInput, task: SnoozeInput): Boolean,
-	dismissCalendarItem(appointment: DismissInput, task: DismissInput): Boolean,
+	snoozeCalendarItem(appointment: [SnoozeInput], task: SnoozeInput): Boolean,
```
silentsakky: should we do similar changes for tasks as well? if not very huge

sachinpatel88: Currently, Reminders are only implemented for appointments and not for Tasks.

PR #319, opened 2020-02-17
PREAPPS-4398 Use Float instead of Int for revision field

silentsakky:


BhavinBathani:
Please resolve code conflict, so that we can move this ahead

PR #332, opened 2020-03-24
PREAPPS-4565: add CreateIdentity and some new attrs

jkester1986:


silentsakky:
@jkester1986 please squash commits and remove merge commits, we are using rebase strategy to reduce number of commits nowadays

src/schema/schema.graphql
```diff
@@ -2808,6 +2814,7 @@ type Mutation {
 		url: String
 		view: FolderView
 	): Folder
+	createIdentity(name: String!, attrs: IdentityAttrsInput!): Identities
```
silentsakky: attrs are required in this line and optional in https://github.com/Zimbra/zm-api-js-client/pull/332/files#diff-28ea553bc23a82875eeae07c5452be22R2121

silentsakky: same issue present in modifyIdentity

yashrajpchavda: I think we can directly use `CreateIdentityInput` here

jkester1986: @silentsakky I am updating this line to be optional instead of required. Should I update it for modifyIdentity as well?

@yashrajpchavda looking at the way the other mutations are written, I don't believe this is correct. `CreateIdentityInput` is used in schema.ts.

silentsakky: yes please update for modifyIdentity as well

src/schema/schema.graphql
```diff
@@ -2645,9 +2656,9 @@ type Query {
 		types: SearchType
 	): SearchResponse
 	getTasks(
-		query: String!,
-		limit: Int!,
-		offset: Int!,
+		query: String!
```
silentsakky, line 2659: @jkester1986 ignore these changes, I have rebased your PR to remove extra commits and ran this file through prettier

PR #335, opened 2020-03-25
PREAPPS-4602: GraphQL error 'upload not found' shows after attaching .ics file; draft mail in Drafts does not contain .ics file

pl12133:
Remove special case for ICS attachments.

### DIFF
--- a/src/utils/normalize-mime-parts.ts
+++ b/src/utils/normalize-mime-parts.ts
@@ -152,7 +152,6 @@ export function normalizeMimeParts(
 				part.contentType !== 'application/pkcs7-mime' &&
 					part.contentType !== 'application/pkcs7-signature' &&
 					part.contentType !== 'application/x-pkcs7-signature' &&
-					!(part.filename && part.filename.endsWith('.ics')) &&
 					(acc[mode] || (acc[mode] = [])).push(processAttachment(part));
 
 				message.attributes = message.attributes || {};


src/utils/normalize-mime-parts.ts
```diff
@@ -152,7 +152,6 @@ export function normalizeMimeParts(
 				part.contentType !== 'application/pkcs7-mime' &&
 					part.contentType !== 'application/pkcs7-signature' &&
 					part.contentType !== 'application/x-pkcs7-signature' &&
-					!(part.filename && part.filename.endsWith('.ics')) &&
```
avadhutprabhudesai: If my memory serves me right, this was removed for a bug where .ics attachment is visible as an attachment in event edit  screen. Could you please check if that issue has resurfaced?

silentsakky: Reverts code from https://github.com/Zimbra/zm-api-js-client/pull/134, most probably original issue will resurface

pl12133: Yes, that is the intention of this ticket. ICS files are attachments and they should be shown the same as other attachments.

This also fixes [PREAPPS-3937: *.ics attachments not shown even when message is not an invitation](https://jira.corp.synacor.com/browse/PREAPPS-3937)

avadhutprabhudesai: But will this code show .ics attachments in 'Calendar -> edit event' screen? If yes, is that the desired behavior?

mogsdad: This won't fix [PREAPPS-3937](https://jira.corp.synacor.com/browse/PREAPPS-3937), because the AC there is that we should suppress ICS attachments on emails that we're handling (as invitations, responses, updates...) The change in #134 overreached.

pl12133: It would have fixed PREAPPS-3937, but we hide the attachments again a second time in the client. That is fixed by https://github.com/ZimbraOS/zm-x-web/pull/2536

PR #338, opened 2020-03-30
PREAPPS-4334: Refactoring Calendar Reminder code

sachinpatel88:


src/schema/notifications.ts
```diff
@@ -42,17 +93,34 @@ function itemsForKey(notification: any, key: string) {
...
-	)[0];
+	const results = Object.keys(data).filter((dataId: string) => {
+		return dataId.indexOf(partialDataId) !== -1 && predicate(dataId);
+	});
+
+	if (returnFirstResult) {
+		return results[0];
+	}
+
+	return results;
```
avadhutprabhudesai, line 112: return returnFirstResult ? results[0] : results;

src/schema/notifications.ts
```diff
@@ -180,6 +249,11 @@ export class ZimbraNotifications {
 		}
 	};
 
+	private handleAppointmentNotifications = (notification: Notification) => {
+		const items = itemsForKey(notification, 'appt');
```
avadhutprabhudesai, line 253: Do we need this variable?

sachinpatel88: Better to have for readability. Also, It's been used in most of the such calls so for consistency as well.

src/schema/notifications.ts
```diff
@@ -1,4 +1,5 @@
 import gql from 'graphql-tag';
+import assign from 'lodash/assign';
```
silentsakky, line 2: lodash-es

src/schema/notifications.ts
```diff
@@ -42,17 +93,34 @@ function itemsForKey(notification: any, key: string) {
...
+
+	if (returnFirstResult) {
+		return results[0];
+	}
+
+	return results;
+}
+
+function utcFromDateString(dateString: String) {
+	const transformedDate = `${dateString.slice(0, 4)}-${dateString.slice(
```
silentsakky, line 116: not sure what is the logic here but better to have something like zm-x-web/src/utils/date/zimbra-date.js

src/schema/notifications.ts
```diff
@@ -214,6 +288,150 @@ export class ZimbraNotifications {
...
+				let eventEndUtc: number;
+				let duration: number;
+
+				if (allDay) {
+					// In case of allDay event, transform `20200401` to `2020-04-01`
+					// and convert it to utc
+					const startDate = get(invitationComponent, 'start.0.date');
+					const endDate = get(invitationComponent, 'end.0.date');
+
+					eventStartUtc = utcFromDateString(startDate);
```
silentsakky, line 325: can we achieve this using moment?

src/schema/notifications.ts
```diff
@@ -214,6 +288,150 @@ export class ZimbraNotifications {
...
+					duration = eventStartUtc - eventEndUtc;
+				} else {
+					eventStartUtc = get(invitationComponent, 'start.0.utc');
+					eventEndUtc = get(invitationComponent, 'end.0.utc');
+					duration = eventStartUtc - eventEndUtc;
+				}
+
+				const dataToWrite: any = {};
+
+				assign(dataToWrite, {
```
silentsakky, line 336: no need to use assign if dataToWrite is already blank

src/schema/notifications.ts
```diff
@@ -214,6 +288,150 @@ export class ZimbraNotifications {
...
+						fragment ${generateFragmentName('appointments')} on CalendarItemHitInfo {
+								id
+								alarm
+								inviteId
+						}
+				`
+				});
+				// If existing event is being updated
+				if (cachedEventDetails) {
+					assign(dataToWrite, cachedEventDetails);
```
silentsakky, line 388: are we overwriting new data with old cache data?

src/schema/notifications.ts
```diff
@@ -214,6 +288,150 @@ export class ZimbraNotifications {
...
+					});
+
+					// Find cached queries based on folder id and update results
+					const query = `(\"inid:\\\\\\"${item.folderId}\\\\\\"\").*(\"types\":\"appointment\")`;
+					const queryRegex = new RegExp(query);
+					const appointmentListKey: any = findDataId(
+						this.cache,
+						'$ROOT_QUERY.getAppointments',
+						dataId => {
+							return queryRegex.test(dataId);
```
silentsakky, line 421: implicit return can be removed

src/schema/notifications.ts
```diff
@@ -478,6 +696,34 @@ export class ZimbraNotifications {
...
 	};
 
+	private writeAppointmentsToCache = (
+		appointmentListKey: any,
+		dataToWrite: any
+	) => {
+		const variables = getVariablesFromDataId(appointmentListKey);
+		const appointmentList: any = this.cache.readQuery({
+			query: AppointmentsQuery,
+			variables
```
silentsakky, line 706: can we have multiple Appointments queries in cache?
if yes then how are we checking updated data belongs to any query or not?

src/schema/notifications.ts
```diff
@@ -33,6 +36,54 @@ const writeNewMailQuery = gql`
 	}
 `;
 
+const AppointmentsQuery = gql`
```
itsmeprashant, line 39: This will get affected if someone changes into cache fragment of query.

itsmeprashant: @silentsakky

src/schema/notifications.ts
itsmeprashant: https://github.com/Zimbra/zm-api-js-client/pull/338/files#r404694170

src/schema/notifications.ts
sachinpatel88: In api-js-client, we don't have lodash-es module. Should I install it?

src/schema/notifications.ts
sachinpatel88: We are not using moment in this repo. Shall we install moment?

I had refused to install because it will be used at very few places.

PR #344, opened 2020-04-08
PREAPPS-4623: Appointments older than particular date disappearing from calendar views and Details screen cannot be viewed

lavegupta:


src/apollo/zimbra-in-memory-cache.ts
```diff
@@ -15,6 +15,8 @@ const dataIdFromPath = (result: any, path: string) => {
 
 const dataIdFromObject = (object: any): string | null | undefined => {
 	switch (object.__typename) {
+		case 'CalendarItemHitInfo':
```
silentsakky, line 18: Great fix !!!
Thanks @lavegupta @itsmeprashant :100:

lavegupta: All credit goes to @itsmeprashant

PR #345, opened 2020-04-17
PREAPPS-4641 Propose New Time for meetings -- Part1

damini-vashishtha:


package-lock.json
```diff
@@ -2106,7 +2106,7 @@
     },
     "@types/valid-url": {
       "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/@types/valid-url/-/valid-url-1.0.2.tgz",
+      "resolved": "http://registry.npmjs.org/@types/valid-url/-/valid-url-1.0.2.tgz",
```
silentsakky, line 2109: not needed changes

src/schema/schema.graphql
```diff
@@ -737,6 +739,11 @@ type TzOnsetInfo { # <standard>, <daylight>
 	sec: Int
 }
 
+type AppointmentInfo { # <appt>; aka AppointmentHitInfo
+id: ID!
```
silentsakky, line 743: indentation

package-lock.json
damini-vashishtha: Done.

BhavinBathani: This is still showing some permission-mod related changes like "100644 → 100755". Can't we get rid of that as well.

src/schema/schema.graphql
```diff
@@ -737,6 +739,11 @@ type TzOnsetInfo { # <standard>, <daylight>
 	sec: Int
 }
 
+type AppointmentInfo { # <appt>; aka AppointmentHitInfo
```
silentsakky, line 742: how is this different then CalendarItemHitInfo ?

silentsakky: both contains same data

damini-vashishtha: CalendarItemHitInfo having so many mandatory fields, which is not the case here

silentsakky: ok

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "@zimbra/api-client",
 	"amdName": "zmApiJsClient",
-	"version": "36.1.0",
+	"version": "36.2.0",
```
silentsakky, line 4: why do we need this?

damini-vashishtha: I am not sure.I just did rebase and It came from there. In blame, showing  Yash wrote this.
In masters also its showing the same version

silentsakky:
Changes moved to https://github.com/Zimbra/zm-api-js-client/pull/359

PR #351, opened 2020-04-27
PREAPPS-4827: Implement GetCustomMetadata and SetCustomMetadata

lavegupta:


src/normalize/entities.ts
```diff
@@ -415,11 +415,21 @@ const ContactAttributes = new Entity({
 	image: ImageFields
 });
 
+const CustomMetaAttributes = new Entity({
+	chatConversationId: 'chatConversationId'
```
silentsakky, line 419: zimlet developers will need flexibility to define their own meta tags, so hardcoding here will not help them

src/normalize/entities.ts
```diff
@@ -415,11 +415,21 @@ const ContactAttributes = new Entity({
...
+	section: 'section',
+	_attrs: ['attributes', CustomMetaAttributes]
+});
+
 const contactFields = {
 	d: 'date',
 	l: 'folderId',
 	rev: 'revision',
 	sf: 'sortField',
+	meta: ['meta', CustomMetadata],
```
silentsakky, line 432: this is not required

src/normalize/entities.ts
```diff
@@ -415,11 +415,21 @@ const ContactAttributes = new Entity({
 	image: ImageFields
 });
 
+const CustomMetaAttributes = new Entity({
+	chatConversationId: 'chatConversationId'
+});
+
+const CustomMetadata = new Entity({
+	section: 'section',
```
silentsakky, line 423: if key and value are same then we don't need to specify it here

src/schema/schema.graphql
```diff
@@ -1388,6 +1402,7 @@ type Contact {
 	memberOf: String
 	attributes: ContactAttributes
 	members: [ContactListMember]
+	meta: [CustomMetadataMeta]
```
silentsakky, line 1405: we can remove this

src/schema/schema.ts
```diff
@@ -410,6 +413,18 @@ export function createZimbraSchema(
 				revokeRights: (_, variables) =>
 					client.revokeRights(variables.input as RevokeRightsInput),
 				revokeTrustedDevice: client.revokeTrustedDevice,
+				setCustomMetadata: (_: any, variables: any) => client
+						.jsonRequest({
```
silentsakky, line 417: we don't do jsonRequest from here directly, it needs to be moved to batch-client

lavegupta: Done

src/schema/schema.graphql
lavegupta: Done

src/normalize/entities.ts
lavegupta: Done

src/normalize/entities.ts
lavegupta: Done

src/utils/normalize-attrs-custommetadata.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export function normalizeCustomMetaDataAttrs(data: any) {
+	let attrs: any = [];
+
+	Object.keys(data._attrs).forEach(
+		key =>
+			typeof data._attrs[key] === 'string' &&
+			attrs.push({ key, value: data._attrs[key] }) &&
+			delete data._attrs[key]
```
silentsakky, line 28: why are we deleting keys here?

lavegupta: Removed

PR #355, opened 2020-05-05
PREAPPS-4860: Appointments with custom recurrence gets disappear

sachinpatel88:


src/schema/schema.ts
```diff
@@ -134,6 +134,12 @@ export function createZimbraSchema(
 				},
 				getAppointments: (_: any, variables) =>
 					client.search(variables as SearchOptions),
+				getReminders: (_: any, variables) => {
+					// `getReminders` query is same as `getAppointments`. Created separate query to
+					// avoid cache conflicts between reminders and appointments components due to same query name.
+					variables.types = 'appointment';
```
silentsakky, line 140: this should be passed by caller, task also can have reminders in future

PR #358, opened 2020-05-07
PREAPPS-4813 : Briefcase | Attach file from Briefcase to email

auminpatel:


src/schema/schema.graphql
```diff
@@ -1734,6 +1734,16 @@ input AttachmentInput {
 	existingAttachments: [ExistingAttachmentInput]
 }
 
+input DocumentIdInput {
+	id: String
```
sachinpatel88, line 1738: Shouldn't it be `ID`?

src/normalize/entities.ts
```diff
@@ -191,6 +191,17 @@ const AttachmentsInfo = new Entity({
...
 });
 
+
+const DocumentIds = new Entity({
+	id : 'id'
+})
+
+const AttachmentsAndDocumentsInfo = new Entity({
+	aid: 'attachmentId',
+	doc: ['documents', DocumentIds] ,
```
sachinpatel88, line 201: I am not sure but Is it required to use normalizer for `DocumentIds`? Anyway its returning id for id.

sachinpatel88:
Branch name typo in PR title: PREAPPS-4183 should be PREAPPS-4813

src/schema/schema.graphql
```diff
@@ -1734,6 +1734,16 @@ input AttachmentInput {
 	existingAttachments: [ExistingAttachmentInput]
 }
 
+input DocumentIdInput {
```
silentsakky, line 1737: we should call it DocumentInput
you never know in future you will be passing more data then just id

src/schema/schema.graphql
```diff
@@ -1783,7 +1793,7 @@ input SendMessageInput {
 	id: ID #id
 	origId: ID #origid
 	folderId: ID # l
-	attach: [AttachmentInput]
+	attach: [AttachmentAndDocumentInput]
```
silentsakky, line 1796: let's not rename it, document is also getting passed as attachment only

src/schema/schema.graphql
auminpatel: Done

src/normalize/entities.ts
auminpatel: as we are returuning id for id i dont think so

src/schema/schema.graphql
auminpatel: done

src/schema/schema.graphql
auminpatel: done

src/normalize/entities.ts
```diff
@@ -186,8 +186,14 @@ const ExistingAttachmentsInfo = new Entity({
 	part: 'part'
 });
 
+
+const DocumentInfo = new Entity({
+	id : 'id'
```
silentsakky, line 191: if key and value are same then it should not belong this file
this file should only be used in case of mapping for short keys to it's longer versions

auminpatel: Done

PR #361, opened 2020-05-12
PREAPPS-4814 : Briefcase | Upload files to briefcase from email

auminpatel:


src/normalize/entities.ts
```diff
@@ -470,12 +470,18 @@ export const Document = new Entity({
 	ct: 'contentType'
 });
 
+export const MessagePartInputForDocuments = new Entity({
```
silentsakky, line 473: can we use MimePartInput ? it looks same

silentsakky: or ExistingAttachmentInput

auminpatel: actually we cant use , as MimePartInput or ExistingAttachmentInput uses messageId as mid for mapping , whereas in SaveDocument it uses id for messageId

auminpatel:
closing this as created new PR

PR #363, opened 2020-05-14
PREAPPS-4759: Tags support for Contacts -- Edit Contact / Assigning tag from list

sachinpatel88:


src/schema/notifications.ts
```diff
@@ -219,13 +219,17 @@ export class ZimbraNotifications {
 
 	private processContactNotifications = (items: any) => {
 		let searchResponse: any = {};
+		// If tag or tagName present in notification item, skip notification updation
```
sachinpatel88, line 222: We're optimistically updating cache on client side.

silentsakky: optimistic update should be ok but we shouldn't ignore updates coming in notifications
consider scenario, user has updated from outlook and expecting that change to be reflected in webapp without reload

sachinpatel88: Ah.. Not considered that case. Thanks. 👍

src/schema/notifications.ts
```diff
@@ -303,21 +329,30 @@ export class ZimbraNotifications {
...
+			// of item. This is required when multiple notification is trying to readfrom cache,
+			// `searchResponse[query].contacts` gets `id` `Contact:id` instead of
+			// `id` for 2nd onwards requests.
 			searchResponse[query].contacts = addNewItemToList(
 				searchResponse[query].contacts,
 				item,
-				sortBy
+				sortBy,
+				(item.tags || item.tagNames) &&
+					this.findTagContact(new RegExp(`^(Contact:)${item.id}+`))
```
sachinpatel88, line 342: change since last approved PR.

PR #365, opened 2020-05-19
PREAPPS-4934: Move notifications handling from api-js to x-web

yashrajpchavda:
- moved the notifications handler to zm-x-web, updated necessary code in api-js for the same pupose
- Improvements in the notifications code to be done in a separate ticket


src/batch-client/index.ts
```diff
@@ -1477,12 +1480,16 @@ export class ZimbraBatchClient {
...
 
-			if (notifications && this.notificationHandler) {
+			if (notifications && this.notificationsEmitter) {
 				// as notification handling happens in synchronous way, if the notifications count is really higher (for bulk operations)
 				// the UI would freeze because of the JavaScript execution time. Hence, delayed the notification handling to give some time
 				// for the render to happen in the JavaScript event loop
 				setTimeout(() => {
-					this.notificationHandler && this.notificationHandler(notifications);
+					this.notificationsEmitter &&
+						this.notificationsEmitter.emit(
```
silentsakky, line 1489: curious, if we are using event emitter then do we need setTimeout wrapping it? as I feel event emitter works on event loop only

src/schema/schema.ts
```diff
@@ -41,7 +41,6 @@ import { ZimbraSchemaOptions } from './types';
 
 import { ZimbraBatchClient } from '../batch-client';
 import { coerceBooleanToString } from '../utils/coerce-boolean';
-import { ZimbraNotifications } from './notifications';
```
silentsakky, line 44: should we delete this file as well?

src/batch-client/index.ts
yashrajpchavda: Yes now the setTimeout would be irrelevant, I will remove it...

src/schema/schema.ts
yashrajpchavda: Yep that was the last step to do, I will just remove it...

yashrajpchavda: Done - deleted

src/batch-client/index.ts
yashrajpchavda: Removed

silentsakky: awesome

PR #369, opened 2020-05-26
PREAPPS-3801: Added input type in DocumentInput and updated normalize…

madhavdhuppe:
… mime part function.

### DIFF
--- a/src/utils/normalize-mime-parts.ts
+++ b/src/utils/normalize-mime-parts.ts
@@ -73,9 +73,10 @@ export function normalizeMimeParts(
 	message: { [key: string]: any },
 	{ origin, jwtToken }: { jwtToken?: string; origin?: string }
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = ({ ...attachment }, updatedContentDisposition='') => {
 		attachment.messageId = attachment.messageId || message.id;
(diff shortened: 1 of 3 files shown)


src/utils/normalize-mime-parts.ts
```diff
@@ -146,13 +147,14 @@ export function normalizeMimeParts(
 
 			// remaining non-body, non-enclosure parts are attachments:
 			if (!isBody && type.split('/')[0] !== 'multipart') {
+				const updateDisposition = part.contentId && 'inline'
```
madhavdhuppe, line 150: If contentId is Present then its an inline attachment but backend is sending disposition  as "attachment" in case document so need to update here
Please let me know if not understood

BhavinBathani: I wonder why this was not needed previously with email-composer.

madhavdhuppe: this case came to known with briefCase images only

itsmeprashant: Are we passing content-disposition as inline when add image from briefcase and trigger save draft request?

itsmeprashant: does this happen with legacy also?

PR #370, opened 2020-05-29
PREAPPS-4567: Selecting an existing persona in Settings > Accounts opens the Persona edit form to modify or delete personas

sachinpatel88:


src/batch-client/index.ts
```diff
@@ -547,6 +548,19 @@ export class ZimbraBatchClient {
 			singleRequest: true
 		}).then(Boolean);
 
+	public deleteIdentity = ({ id, name }: DeleteIdentityInput) =>
+		this.jsonRequest({
+			name: 'DeleteIdentity',
+			namespace: Namespace.Account,
+			body: {
+				identity: {
```
madhavdhuppe, line 556: can we write something like this,
**public deleteIdentity = (identity: DeleteIdentityInput) =>**

then in Body
**identity: identity**

silentsakky: ideally instead of passing name and id separately we should have passed identity object itself, but since this issue is with createIdentity as well
so either we change for all or don't change

sachinpatel88: Yes, with same concern I kept it as it is. I tried and tested changing all the queries (`createIdentity` and `modifyIdentity`) along with `deleteIdentity` as generic as possible.

PR #371, opened 2020-06-02
PREAPPS-4971: Fix reminders refetch only when notifications for appt present

yashrajpchavda:
- updated noop code to include parameters

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -3,7 +3,6 @@ import castArray from 'lodash/castArray';
 import get from 'lodash/get';
 import isError from 'lodash/isError';
 import mapValues from 'lodash/mapValues';
-import emitter from 'mitt';
 import { denormalize, normalize } from '../normalize';
 import {
 	AccountRights,
@@ -132,6 +131,7 @@ import {
 	GetSMimePublicCertsOptions,
 	LoginOptions,
 	ModifyProfileImageOptions,
+	NoOpOptions,
 	RecoverAccountOptions,
 	RelatedContactsOptions,
 	ResetPasswordOptions,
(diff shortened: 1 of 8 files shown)


src/batch-client/index.ts
```diff
@@ -1704,7 +1706,10 @@ export class ZimbraBatchClient {
 			this.sessionId ||
 			(this.sessionHandler && this.sessionHandler.readSessionId()),
 		origin: this.origin,
-		userAgent: this.userAgent
+		userAgent: this.userAgent,
+		...(this.notifier.getSequenceNumber() && {
+			sessionSeq: this.notifier.getSequenceNumber()
```
silentsakky, line 1711: are we maintaining same seq number when request is fired through old zimbra-client?

src/request/index.ts
```diff
@@ -247,14 +247,23 @@ export function jsonRequest(
 		[soapRequestName]: soapCommandBody(options)
 	};
 
+	let fetchOptions;
+
+	if (requestOptions.fetchOptions) {
+		fetchOptions = {
+			signal: requestOptions.fetchOptions.signal
```
silentsakky, line 254: what is signal?

silentsakky: can't we just send it as requestOptions.signal? instead of adding fetchOptions layer?

silentsakky: its for abortcontroller, so ignore above comments

src/batch-client/index.ts
yashrajpchavda: Ah, had missed that. The answer is, no we are not. I just checked that only `csrf` is added in old client, it is not adding `userAgent`, `sessionId`, `sessionSeq`, `accountName`, and `jwt`. Should we add them too?

PR #376, opened 2020-06-10
PREAPPS-4568-1: Added persona folder and replyForward fields

sachinpatel88:


sachinpatel88:
Have updated git commit message config recently for api-js-client on my local. Next time it will reflect in expected format. Please ignore this commit message.

src/batch-client/index.ts
```diff
@@ -192,6 +192,18 @@ function updateAbsoluteFolderPath(
...
+ * Some fields expect values to be String if it's single value, else array of values.
+ * i.e. Persona Email addresses - if single email address, it should be 'email@address.com',
+ * else it should be ['email1@address.com', 'email2@address.com']
+ * This method will return Array in all cases by converting string to Array of String.
+ * To empty value, it's value should be '' and not [].
+ * @param value Array or null value
+ */
+function convertStringAndArrayValues(value: any) {
+	const result = [].concat(value).filter(Boolean);
+	return result.length === 0 ? '' : result;
```
lavegupta, line 205: `return result.length ? result : '';`

src/batch-client/index.ts
```diff
@@ -192,6 +192,16 @@ function updateAbsoluteFolderPath(
 		return folder;
 	});
 }
+/**
+ * Return an empty string in case it's empty array or null value,
+ * else return an Array.
+ *
+ * @param value An Array or empty String
```
silentsakky, line 199: graphql would expect single data type to be returned, how are we handling case when we return string instead of array?

src/batch-client/index.ts
```diff
@@ -841,7 +890,37 @@ export class ZimbraBatchClient {
...
+						zimbraPrefWhenInFolderIds,
+						zimbraPrefWhenSentToAddresses,
+						...restAttrs
+					},
+					...restIdentity
+				}: any) => ({
+					...restIdentity,
+					_attrs: {
+						...restAttrs,
+						zimbraPrefWhenInFolderIds: []
```
silentsakky, line 906: convertStringAndArrayValues

src/batch-client/index.ts
```diff
@@ -192,6 +192,16 @@ function updateAbsoluteFolderPath(
...
 }
+/**
+ * Return an empty string in case it's empty array or null value,
+ * else return an Array.
+ *
+ * @param value An Array or empty String
+ */
+function convertStringAndArrayValues(value: any) {
+	const result = [].concat(value).filter(Boolean);
+	return result.length ? result : '';
```
silentsakky, line 203: i think graphql will expect an empty array instead of empty string

src/batch-client/index.ts
sachinpatel88: We convert it only when we send data to Server. I have updated code comments. We are doing so, because, for Server, following values means:

`[]` - No change to reflect on server.
`''` - Set value to empty for field.

src/batch-client/index.ts
sachinpatel88: For `getIdentities`, On client side, we need arrays only. So not converting using this method. Added code comment.

src/batch-client/index.ts
sachinpatel88: Covered by https://github.com/Zimbra/zm-api-js-client/pull/376/files#r444237374

src/batch-client/index.ts
sachinpatel88: Added comment as discussed.

src/batch-client/index.ts
```diff
@@ -473,11 +500,51 @@ export class ZimbraBatchClient {
...
+							...(zimbraPrefWhenSentToAddresses && {
+								zimbraPrefWhenSentToAddresses: []
+									.concat(zimbraPrefWhenSentToAddresses)
+									.filter(Boolean)
+							}),
+							...(zimbraPrefWhenInFolderIds && {
+								zimbraPrefWhenInFolderIds: []
+									.concat(zimbraPrefWhenInFolderIds)
+									.filter(Boolean)
+							})
```
sachinpatel88, line 542: Updated. Tested and it's working fine.

PR #377, opened 2020-06-11
PREAPPS-4972: Updated fields into the ExternalAccountModifyAttrsInput as per required 

madhavdhuppe:
-Added password field into the ExternalAccountModifyAttrsInput and remove emailAddress as it was causing issues.
-Added attributes in ExternalAccountModifyAttrsInput.

### DIFF
--- a/src/schema/schema.graphql
+++ b/src/schema/schema.graphql
@@ -1271,6 +1271,7 @@ type DataSource {
 	isEnabled: Boolean
 	name: String
 	pollingInterval: Float
+	leaveOnServer: Boolean
 	port: String
 	replyToAddress: String
 	replyToDisplay: String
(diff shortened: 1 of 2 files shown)


src/schema/schema.graphql
```diff
@@ -2580,7 +2580,6 @@ input ExternalAccountModifyAttrsInput {
 	accountType: AccountType
 	defaultSignature: ID
 	description: String
-	emailAddress: String
```
madhavdhuppe, line 2583: upon sending emailAddress to backend it was throwing the "Data source already exists" error
so removed this

itsmeprashant: will this change have any impact on add new account?

madhavdhuppe: nope we are modifying "ExternalAccountModifyAttrsInput" only

PR #379, opened 2020-06-16
PREAPPS-4821 : Adding previewpane attribute for briefcase

auminpatel:


src/batch-client/index.ts
```diff
@@ -1573,6 +1579,33 @@ export class ZimbraBatchClient {
...
 		});
+	
+
+		private downloadDoc = ({ id, url ,isSecure}: any) =>
+		fetch(
+			`${this.origin}${url}`,
+			{
+				headers: {
+					...(isSecure && { 'X-Zimbra-Encoding': 'x-base64' }),
+					...(this.csrfToken && {
```
auminpatel, line 1590: have updated headers to this
`{
				...(this.csrfToken && {headers: {
						'X-Zimbra-Csrf-Token': this.csrfToken
					}
				}),
				credentials: 'include'
			}`

will push the code  - once reviewed ( have ran automation recently)

src/batch-client/index.ts
```diff
@@ -1573,6 +1579,31 @@ export class ZimbraBatchClient {
 				});
 			}
 		});
+	
+
+		private downloadDoc = ({ id, url }: any) =>
```
silentsakky, line 1584: indentation

silentsakky: how is this different then download method? only url?

silentsakky: another weird thing I found downloadMessage passes isSecure as second argument, ideally it should be third argument, is this regression of the functionality?

src/schema/schema.graphql
```diff
@@ -223,7 +223,7 @@ enum NeedIsMemberType {
 	none
 }
 
-enum ReadingPaneLocation {
+enum PaneLocation {
```
silentsakky, line 226: I think previous name was proper

src/batch-client/index.ts
auminpatel: yes , only url structure is changing , as for messages we need to send part and we are generating url here ,
for documents i made the url and passed it , 
we can either  url for all methods and pass it  and keep a common function.

havent checked the downloadMessage functionality will have to check it

src/batch-client/index.ts
```diff
@@ -603,13 +603,19 @@ export class ZimbraBatchClient {
 		}).then(Boolean);
 
 	public downloadAttachment = ({ id, part }: any) =>
-		this.download({ id, part }).then(({ id, part, content }: any) => ({
+		this.download({ id, part, url:`/service/home/~/?auth=co&id=${id}&part=${part}` }).then(({ id, part, content }: any) => ({
```
silentsakky, line 606: do we need to pass part? is it getting used in download method?

silentsakky: I think download method can just return content and addition of id and part with that response can be handled in wrapper functions downloadAttachment and downloadDocument

src/schema/schema.graphql
auminpatel: Done

src/batch-client/index.ts
auminpatel: done , just send isSecure (were required) and url and receiving  content

src/batch-client/index.ts
```diff
@@ -603,13 +603,19 @@ export class ZimbraBatchClient {
 		}).then(Boolean);
 
 	public downloadAttachment = ({ id, part }: any) =>
-		this.download({ id, part }).then(({ id, part, content }: any) => ({
+		this.download({ url:`/service/home/~/?auth=co&id=${id}&part=${part}` }).then(({ id, part, content }: any) => ({
```
silentsakky, line 606: callback in then expects id, part will be passed by download function but we have removed it

src/batch-client/index.ts
```diff
@@ -603,13 +603,19 @@ export class ZimbraBatchClient {
...
 
+	public downloadDocument = ({ id, url }: any) =>
+		this.download({ url }).then(({ content }: any) => ({
+			id: id,
+			content
+		}));
+
 	public downloadMessage = ({ id, isSecure }: any) =>
-		this.download({ id, isSecure }).then(({ id, content }: any) => ({
+		this.download({ isSecure , url:`/service/home/~/?auth=co&id=${id}` }).then(({ id, content }: any) => ({
```
silentsakky, line 618: same as https://github.com/Zimbra/zm-api-js-client/pull/379/files#r445124037

src/batch-client/index.ts
auminpatel: Done

src/batch-client/index.ts
auminpatel: Done

PR #380, opened 2020-06-19
PREAPPS-4846 : Adding Device status related graphql queries.

BhavinBathani:
There are one query and three mutations requires to be added.
API - GetDeviceStatusRequest
API - SuspendDeviceRequest
API - ResumeDeviceRequest
API - RemoveDeviceRequest

These will enable us to retrieve necessary data and update the same.

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -856,6 +856,12 @@ export class ZimbraBatchClient {
 			name: 'GetDataSources'
 		}).then(res => mapValuesDeep(res, coerceStringToBoolean));
 
+	public getDeviceStatus = () =>
+		this.jsonRequest({
+			name: 'GetDeviceStatus',
+			namespace: Namespace.Sync
+		}).then(res => get(res, 'device', []));
+
 	public getFilterRules = () =>
 		this.jsonRequest({
 			name: 'GetFilterRules'
@@ -1373,6 +1379,17 @@ export class ZimbraBatchClient {
 			}
 		}).then(resp => resp.relatedContacts.relatedContact);
 
+	public removeDeviceSync = (deviceId: String) =>
+		this.jsonRequest({
+			name: 'RemoveDevice',
+			namespace: Namespace.Sync,
+			body: {
+				device: {
+					id: deviceId
+				}
+			}
+		});
+
 	public resetPassword = ({ password }: ResetPasswordOptions) =>
 		this.jsonRequest({
 			name: 'ResetPassword',
@@ -1384,6 +1401,17 @@ export class ZimbraBatchClient {
 		}).then(() => true);
 
 	public resolve = (path: string) => `${this.origin}${path}`;
+
+	public resumeDeviceSync = (deviceId: String) =>
+		this.jsonRequest({
+			name: 'ResumeDevice',
+			namespace: Namespace.Sync,
+			body: {
+				device: {
+					id: deviceId
+				}
+			}
+		});
 
 	public revokeAppSpecificPassword = (appName: string) =>
 		this.jsonRequest({
@@ -1543,6 +1571,17 @@ export class ZimbraBatchClient {
 			},
 			singleRequest: true
 		}).then(Boolean);
+
+	public suspendDeviceSync = (deviceId: String) =>
+		this.jsonRequest({
+			name: 'SuspendDevice',
+			namespace: Namespace.Sync,
+			body: {
+				device: {
+					id: deviceId
+				}
+			}
+		});
 
 	public taskFolders = () =>
 		this.jsonRequest({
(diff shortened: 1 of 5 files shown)


src/schema/schema.graphql
```diff
@@ -3136,6 +3150,9 @@ type Mutation {
 	): Boolean
 	createTag(tag: CreateTagInput): Tag
 	tagAction(action: FolderActionInput): Boolean
+	suspendDeviceSyncMutation(deviceId: String) : Device
```
yashrajpchavda, line 3153: can remove the word `Mutation` from these names

src/batch-client/index.ts
```diff
@@ -1529,6 +1557,17 @@ export class ZimbraBatchClient {
...
 		}).then(Boolean);
 
+	public suspendDeviceSync = (deviceId: String) =>
+		this.jsonRequest({
+			name: 'SuspendDevice',
+			namespace: Namespace.Sync,
+			body: {
+				device: {
+					id: deviceId
+				}
```
madhavdhuppe, line 1567: can we create type for this as its getting reused many places

src/schema/generated-schema-types.ts
```diff
@@ -3368,24 +3404,24 @@ export type QueryShareInfoArgs = {
...
-  recip?: Maybe<Scalars['Int']>,
-  sortBy?: Maybe<SortBy>,
-  types?: Maybe<SearchType>,
-  resultMode?: Maybe<Scalars['String']>,
-  inDumpster?: Maybe<Scalars['Boolean']>
-};
-
-
-/** 
+  contact?: Maybe<Scalars['String']>;
```
madhavdhuppe, line 3407: i think you run into conflicts due to this

src/batch-client/index.ts
```diff
@@ -848,6 +848,12 @@ export class ZimbraBatchClient {
 			name: 'GetDataSources'
 		}).then(res => mapValuesDeep(res, coerceStringToBoolean));
 
+	public getDeviceStatus = () =>
+		this.jsonRequest({
+			name: 'GetDeviceStatus',
+			namespace: Namespace.Sync
+		}).then(res => get(res, 'device', []));
```
silentsakky, line 855: get with third argument doesn't work in case of graphql data

src/schema/schema.graphql
```diff
@@ -2799,6 +2799,19 @@ type GetTrustedDevicesResponse {
...
 
+type Device {
+	id: ID
+	name: String
+	firstReqReceived: Int
+	lastPolicyUpdate: Int
+	lastUsedDate: String
+	protocol: Float
+	provisionable: Boolean
+	status: Int
```
silentsakky, line 2810: we should have enum for this

src/schema/schema.ts
```diff
@@ -447,7 +448,16 @@ export function createZimbraSchema(
 				modifyWhiteBlackList: (_, { whiteBlackList }) =>
 					client.modifyWhiteBlackList(whiteBlackList as WhiteBlackListInput),
 				createTag: (_, { tag }) => client.createTag(tag as CreateTagInput),
-				tagAction: (_, { action }) => client.action(ActionType.tag, action)
+				tagAction: (_, { action }) => client.action(ActionType.tag, action),
+				suspendDeviceSync: (_, { deviceId }) => {
```
silentsakky, line 452: as far as I know all queries/mutations invoked from user-sessions-management zimlet is directly hitting graphql instead of going from this layer

src/batch-client/index.ts
BhavinBathani: It will help us if there will be more than one params, here its id only. lets keep this as it is.

src/schema/schema.ts
BhavinBathani: I talked with Rupali and confirm the same by checking zm-mailbox/soap but there is no @graphql annotation at all.
This has no support from back-end. we need this layer.

src/schema/generated-schema-types.ts
BhavinBathani: Wish me luck then ;)

src/schema/schema.graphql
BhavinBathani: I had tried but failed to do so as we are getting Byte/numbers from server, keeping this as it is.

src/batch-client/index.ts
BhavinBathani: It is being used at other places as well. If its not harmful then let me keep this.

silentsakky: yeah but it will not work when you get null as response

BhavinBathani: Fixed.

PR #382, opened 2020-06-22
PREAPPS-4907 Plaintext email support from Settings

damini-vashishtha:


src/schema/schema.graphql
```diff
@@ -1325,6 +1325,7 @@ type Preferences {
 	zimbraPrefCalendarShowPastDueReminders: Boolean
 	zimbraPrefCalendarToasterEnabled: Boolean
 	zimbraPrefComposeDirection: String
+	zimbraPrefComposeFormat: String
```
silentsakky, line 1328: I think this should be an enum

PR #383, opened 2020-06-24
PREAPPS-4829: Added value in folderView enum.

madhavdhuppe:


madhavdhuppe:
> Having commit message like below doesn't add value
> `PREAPPS-4829: Added value in folderView enum.`
> can we add more details on why this was needed and what it does?

Updated

PR #385, opened 2020-06-29
Preapps 4972 1: Change password + settings of an external account

madhavdhuppe:


src/schema/schema.graphql
```diff
@@ -2646,6 +2647,8 @@ input ExternalAccountModifyAttrsInput {
 	username: String
 	password: String
 	host: String
+	l: ID! # TODO: Normalize to `folderId`
```
itsmeprashant, line 2650: folderId

madhavdhuppe: To update this we have to make changes at many places in ZM-x-web and test them as well can we skip this for now 
i will do this work whenever i get free time

yashrajpchavda: I'd like this to be done in this ticket itself, just don't want to create any techdebt to be fixed later.

By the way, if I understand correctly, it just needs a normalization entity to be created and used while invoking API, right? Or do we need anything else also?

madhavdhuppe: Actually yash whole datasource related API's are already using 'l' instead of 'folderId'.
so, i had to use that way only 
As discussed creating a task ticket for this.
CC @silentsakky @itsmeprashant

PR #395, opened 2020-07-15
PREAPPS-4979-2 register the handler for refresh event

damini-vashishtha:


src/batch-client/notifier/index.ts
```diff
@@ -2,7 +2,7 @@ import emitter from 'mitt';
 
 export class Notifier {
 	private emitter: any;
-	private eventName: string = 'notify';
+	private eventName = { notify: 'notify', refresh: 'refresh' };
```
yashrajpchavda, line 5: lets rename this to `events`

damini-vashishtha: Done

PR #400, opened 2020-08-03
PREAPPS-4517:- TypeError when opening Synacor Global Address List

madhavdhuppe:
the reason we need to cast userCertificate value to string:-

1) User Certificate comes as **string** when their is only one certificate available :-
image

2) User certificate comes as **Array** when their are multiple certificates:-
image

### DIFF
--- a/src/utils/normalize-otherAttribute-contact.ts
+++ b/src/utils/normalize-otherAttribute-contact.ts
@@ -117,6 +117,15 @@ export function normalizeOtherAttr(data: any) {
 				: castValueOfNickname;
 		}
 
+		let castValueOfUserCert: any = contact._attrs['userCertificate'];
+
+		if (castValueOfUserCert) {
+			// To Do:- Handle multiple UserCertificates for a contact
+			contact._attrs['userCertificate'] = Array.isArray(castValueOfUserCert)
+				? castValueOfUserCert[0]
+				: castValueOfUserCert;
+		}
+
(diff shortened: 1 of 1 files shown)


src/schema/custom-scalar.ts
```diff
@@ -0,0 +1,28 @@
...
+	serialize(value) {
+		if (typeof value !== 'string' && typeof value !== 'object') {
+			throw new Error('Value must be either a String or an Array');
+		}
+
+		if (typeof value === 'object' && typeof value.length !== 'number') {
+			throw new Error('Value must be either a String or an Array');
+		}
+
+		return value;
```
auminpatel, line 15: can we write it like this  : 
`if (typeof value !== 'string' && typeof value !== 'object' || typeof value === 'object' && typeof value.length !== 'number') `

src/schema/schema.graphql
```diff
@@ -1610,7 +1612,7 @@ type ContactAttributes {
 	notes: String
 	thumbnailPhoto: String
 	image: ContactImage
-	userCertificate: String
+	userCertificate: StringOrArray
```
silentsakky, line 1615: graphql is a strict typed language, we shouldn't have a property whose type is unknown at build time

silentsakky: why can't we always convert this to an array?

madhavdhuppe: ok

PR #410, opened 2020-09-24
PREAPPS-5327: Intermittent | Application doesn't load after browser refresh, Error: t.props.refetchAppointments is not a function

sachinpatel88:


src/batch-client/index.ts
```diff
@@ -1789,7 +1789,7 @@ export class ZimbraBatchClient {
 			(this.sessionHandler && this.sessionHandler.readSessionId()),
 		origin: this.origin,
 		userAgent: this.userAgent,
-		...(this.notifier.getSequenceNumber() && {
+		...(typeof this.notifier.getSequenceNumber() !== 'undefined' && {
```
silentsakky, line 1792: why this check?

sachinpatel88: When we reset `sessionSeq` to 0, it's value will be 0 and In that case it will be `false` here and won't pass `sessionSeq`.
So added this check to avoid above mentioned scenario.

silentsakky: this will always return some numeric value right?

sachinpatel88: Yes. And it mostly it gets it's value from [response headers](https://github.com/Zimbra/zm-api-js-client/blob/master/src/batch-client/notifier/index.ts#L47) (`header.context.notify.0`). It is possible that `header.context.notify.0` gets `undefined`. But we set `this.sequenceNo` if header sequence is not `undefined`.

silentsakky (approved):

src/batch-client/notifier/index.ts
```diff
@@ -2,12 +2,15 @@ import emitter from 'mitt';
...
-	private sequenceNo: number = 0;
+	private events: any;
+	private processedSequences: any;
+	private sequenceNo: number;
 
 	constructor() {
 		this.emitter = new (emitter as any)();
+		this.events = { notify: 'notify', refresh: 'refresh' };
+		this.processedSequences = new Map();
+		this.sequenceNo = 0;
```
silentsakky, line 13: maybe we should revert this now, as it's same as earlier

damini-vashishtha (approved):

madhavdhuppe (approved):

yashrajpchavda (approved):

silentsakky (approved):

PR #411, opened 2020-09-30
PREAPPS-5189 : Adding documentAction for handling sharing of documents

auminpatel:


lavegupta (approved):

itsmeprashant (approved):

src/batch-client/index.ts
```diff
@@ -314,6 +315,21 @@ export class ZimbraBatchClient {
...
 	};
 
+	public documentActionresponse = (type: ActionType, options: ActionOptions) => {
+		const { ids, id, ...rest } = options;
+
+		return this.jsonRequest({
+			name: type,
+			body: {
+				action: {
+					id: id || (ids || []).join(','),
```
silentsakky, line 325: do we need two separate parameters?
`['single'].join(',') ==> 'single'`

auminpatel: that works, 
Done

src/batch-client/index.ts
```diff
@@ -314,6 +315,21 @@ export class ZimbraBatchClient {
...
 	};
 
+	public documentActionresponse = (type: ActionType, options: ActionOptions) => {
+		const { ids, id, ...rest } = options;
+
+		return this.jsonRequest({
+			name: type,
+			body: {
+				action: {
+					id: id || [ids].join(','),
```
silentsakky, line 325: do we need two parameters id and ids?

auminpatel: not required

auminpatel: as this is only used for documentAction response so i have used only id so removed it

src/batch-client/index.ts
```diff
@@ -306,14 +307,29 @@ export class ZimbraBatchClient {
 			name: type,
 			body: {
 				action: {
-					id: id || (ids || []).join(','),
+					id: id || [ids].join(','),
```
silentsakky, line 310: do we need two params ? id and ids?

auminpatel: Many of the actions are passing ids or id, this part is been use in msgaction, Folderaction itemaction so I haven't checked those mutation

auminpatel: Screen Shot 2020-10-06 at 7 03 21 AM
just a example

silentsakky: ok

silentsakky (approved):

PR #432, opened 2020-12-11
ZAPPD-45: Desktop App Sync feature

yashrajpchavda:
PREAPPS-5653: added generated schema file

PREAPPS-5653 fixed uglifyjs error for windows build

PREAPPS-5653: added base64 input in the mimepart

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
 	"license": "BSD-3-Clause",
 	"dependencies": {
 		"@apollo/client": "^3.2.1",
+		"cross-fetch": "^3.0.6",
 		"dataloader": "^1.4.0",
 		"graphql": "^15.3.0",
 		"graphql-tools": "^7.0.4",
--- a/package.json
+++ b/package.json
@@ -80,6 +80,6 @@
 		"tslint-config-prettier": "^1.18.0",
 		"tslint-plugin-prettier": "^2.3.0",
 		"typescript": "^4.0.3",
-		"uglifyjs": "^2.4.11"
+		"uglify-js": "^3.12.1"
 	}
 }
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
 	"license": "BSD-3-Clause",
 	"dependencies": {
 		"@apollo/client": "^3.2.1",
-		"cross-fetch": "^3.0.6",
 		"dataloader": "^1.4.0",
 		"graphql": "^15.3.0",
 		"graphql-tools": "^7.0.4",
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -306,12 +306,14 @@ export class ZimbraBatchClient {
 	private jwtToken?: string;
 	private sessionHandler?: SessionHandler;
 	private userAgent?: {};
+	private authToken?: string;
 
 	constructor(options: ZimbraClientOptions = {}) {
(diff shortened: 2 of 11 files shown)


src/request/index.ts
```diff
@@ -243,6 +248,10 @@ export function jsonRequest(
 		header.context.csrfToken = requestOptions.csrfToken;
 	}
 
+	if (requestOptions.authToken) {
```
silentsakky, line 251: we should add a comment about in which condition this will be used and why we need to do it here

src/batch-client/types.ts
```diff
@@ -61,10 +61,12 @@ export type SessionHandler = {
 };
 
 export interface ZimbraClientOptions {
+	authToken?: string;
 	csrfToken?: string;
 	jwtToken?: string;
 	localStoreClient?: any;
 	notificationHandler?: NotificationHandler;
+	requestAPI?: any;
```
silentsakky, line 69: customFetch

src/schema/schema.graphql
```diff
@@ -3253,6 +3258,7 @@ type Mutation {
 		name: String!
 		parentFolderId: ID
 		url: String
+		isLocalFolder: Boolean
```
silentsakky, line 3261: isLocal?

package.json
```diff
@@ -80,6 +80,6 @@
 		"tslint-config-prettier": "^1.18.0",
 		"tslint-plugin-prettier": "^2.3.0",
 		"typescript": "^4.0.3",
-		"uglifyjs": "^2.4.11"
+		"uglify-js": "^3.12.1"
```
silentsakky, line 83: any specific reason to upgrade this?

src/batch-client/index.ts
```diff
@@ -310,6 +313,10 @@ export class ZimbraBatchClient {
 			batch: false,
 			cache: false
 		});
+
+		if (options.requestAPI) {
```
silentsakky, line 317: we should update readme with this option, so devs should know they can pass custom fetch function here

src/request/index.ts
```diff
@@ -255,7 +264,9 @@ export function jsonRequest(
 		};
 	}
 
-	return fetch(url, {
+	// Use received `requestAPI` passed as params instead of default fetch API
+	// to make `jsonRequest` method compatible with unsupported platforms, i.e. node.js
+	return (requestAPI || fetch)(url, {
```
silentsakky, line 269: there may be some other places also where we are using fetch directly

package.json
silentsakky: although it's always good to upgrade libraries, was just curious

sachinpatel88: It's been deprecated: https://www.npmjs.com/package/uglifyjs

src/schema/schema.graphql
sachinpatel88: As discussed, Folder and mailItems has different key names to identify whether it's local or not. Also, it is extensively used in zm-x-web. Keeping it as it is.

src/batch-client/index.ts
sachinpatel88: Please have a look: https://github.com/Zimbra/zm-api-js-client/tree/PREAPPS-6008#provide-custom-fetch-api

silentsakky (approved):

silentsakky (approved):

PR #440, opened 2021-01-20
ZCS-10127: Do not display external images in Junk even with zimbraPrefDisplayExternalImages=TRUE

madhavdhuppe:


src/batch-client/index.ts
```diff
@@ -1335,16 +1335,27 @@ export class ZimbraBatchClient {
...
+			zimbraPrefMailTrustedSenderList.length === 0) {
+				attrs.zimbraPrefMailTrustedSenderList = '' ;
+		}
+
+		return this.jsonRequest({
 			name: 'ModifyPrefs',
 			namespace: Namespace.Account,
 			body: {
-				_attrs: mapValuesDeep(prefs, coerceBooleanToString)
+				_attrs: attrs
```
silentsakky, line 1353: previously we were doing coerceBooleanToString and now we are doing coerceStringToBoolean

src/batch-client/index.ts
```diff
@@ -1335,16 +1335,27 @@ export class ZimbraBatchClient {
...
 		});
 
-	public modifyPrefs = (prefs: PreferencesInput) =>
-		this.jsonRequest({
+	public modifyPrefs = (prefs: PreferencesInput) => {
+
+		let attrs: any = mapValuesDeep(prefs, coerceStringToBoolean);
+		const {zimbraPrefMailTrustedSenderList} = prefs;
+
+		if (zimbraPrefMailTrustedSenderList &&
```
silentsakky, line 1343: this needs to be handled in a generic way for all array type prefs

madhavdhuppe: sure i will add code for generic way

src/batch-client/index.ts
madhavdhuppe: will update

silentsakky (approved):

src/batch-client/index.ts
```diff
@@ -1335,15 +1341,26 @@ export class ZimbraBatchClient {
 			singleRequest: true
 		});
 
-	public modifyPrefs = (prefs: PreferencesInput) =>
-		this.jsonRequest({
+	public modifyPrefs = (prefs: PreferencesInput) => {
+		let attrs: any = mapValuesDeep(prefs, coerceBooleanToString);
+		Object.keys(attrs).map(pref => {
```
aayushvijayvergiya, line 1346: If there is only the need of iteration of the array, we can use forEach instead of map. 
And we can also skip `Object.keys()` by using for...in loop for the object.

src/batch-client/index.ts
```diff
@@ -1083,10 +1084,15 @@ export class ZimbraBatchClient {
 			namespace: Namespace.Account
 		}).then(res => {
 			let prefs: any = mapValuesDeep(res._attrs, coerceStringToBoolean);
-			prefs.zimbraPrefMailTrustedSenderList =
-				typeof prefs.zimbraPrefMailTrustedSenderList === 'string'
-					? castArray(prefs.zimbraPrefMailTrustedSenderList)
-					: prefs.zimbraPrefMailTrustedSenderList;
+
+			Object.keys(prefs).map(pref => {
```
aayushvijayvergiya, line 1088: If there is only the need of iteration of the array, we can use forEach instead of map.
And we can also skip Object.keys() by using for...in loop for the object.

madhavdhuppe: Object is object not array

src/batch-client/index.ts
madhavdhuppe: Object is object not array

madhavdhuppe: updated

src/batch-client/index.ts
madhavdhuppe: updated

aayushvijayvergiya (approved):

PR #444, opened 2021-01-29
PREAPPS-5772: Support 2FA and forgot password at the same time. | Modern UI

aayushvijayvergiya:
Modified Reset Password request to support password rules

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -1474,16 +1474,18 @@ export class ZimbraBatchClient {
 			}
 		});
 
-	public resetPassword = ({ password, dryRun }: ResetPasswordOptions) =>
+	public resetPassword = ({ password, dryRun, getPasswordRules, cancelResetPassword }: ResetPasswordOptions) =>
 		this.jsonRequest({
 			name: 'ResetPassword',
 			namespace: Namespace.Account,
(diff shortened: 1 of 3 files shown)


src/batch-client/index.ts
```diff
@@ -1457,16 +1457,18 @@ export class ZimbraBatchClient {
...
 			namespace: Namespace.Account,
 			body: {
 				password,
-				dryRun
+				dryRun,
+				getPasswordRules,
+				cancelResetPassword
 			},
 			singleRequest: true
-		}).then(() => true);
```
madhavdhuppe, line 1469: why did you removed this ?

itsmeprashant (approved):

madhavdhuppe (approved):

src/batch-client/index.ts
itsmeprashant: here we were returning a boolean value instead of actual response of server and Aayush changed that.

aayushvijayvergiya: earlier the api wasnt giving any response, so we were explicitly returning boolean, but now we have response, so that's why `then` block is removed.
for reset password case, we wont be passing `getPasswordRules`. Just the password attribute

PR #447, opened 2021-02-22
PREAPPS-5503: adding the unreadDescendent to check if any subfolder or the folder itself has a unread mails in the hireracy

auminpatel:
Function setUnreadDescendentFlag will check only if the folder View is null or Message , null because the follders that are through imap or pop
hasUnreadDescendent will set the flag unreadDescendent if a folders subfolder or itself has a unread count

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -4,6 +4,7 @@ import get from 'lodash/get';
 import isError from 'lodash/isError';
 import mapValues from 'lodash/mapValues';
 import { denormalize, normalize } from '../normalize';
+import { USER_FOLDER_IDS } from './constants'
 import {
 	AccountRights,
 	ActionOptions as ActionOptionsEntity,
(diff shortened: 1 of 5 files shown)


src/batch-client/index.ts
```diff
@@ -169,6 +169,49 @@ function normalizeMessage(
 	);
 }
 
+const hasUnreadDescendent = (folder: any) => {
+	if (
+		folder[
+			folder.name === 'Drafts' || folder.name === 'Outbox'
```
silentsakky, line 175: instead of checking folder names, can we check folder ids?

src/batch-client/index.ts
```diff
@@ -169,6 +169,49 @@ function normalizeMessage(
...
+				? 'nonFolderItemCount'
+				: 'unread'
+		] > 0
+	) {
+		return true;
+	}
+
+	const folderArray = get(folder, 'folders') || [];
+	for (let i = 0; i < folderArray.length; i++) {
```
silentsakky, line 184: one small optimization to not count length with every iteration
`for (let i = 0, len = folderArray.length; i < len; i++) {`

src/batch-client/index.ts
```diff
@@ -169,6 +169,49 @@ function normalizeMessage(
...
+		if (hasUnreadDescendent(folderArray[i])) {
+			return true;
+		}
```
silentsakky, line 187: `return hasUnreadDescendent(folderArray[i]);`

src/batch-client/index.ts
```diff
@@ -169,6 +169,49 @@ function normalizeMessage(
...
+	if (!view || view === FolderView.Message) {
+		folder = {
+			...folder,
+			unreadDescendent: hasUnreadDescendent(folder)
+		};
+
+		if (folderArray) {
+			folder = {
+				...folder,
+				folders: folderArray.map(setUnreadDescendentFlag)
```
silentsakky, line 208: hasUnreadDescendent is already doing a loop on subfolders, here we are also doing same.
can we reduce looping required to achieve end result?

src/batch-client/index.ts
auminpatel: I couldnt find a folderId constant for Outbox so wll keep this, name only

src/batch-client/index.ts
auminpatel: Done

src/batch-client/index.ts
auminpatel: Done

src/batch-client/index.ts
auminpatel: Have tried,but couldnt find a solution ,to reduce looping , for certian level when a unreadDescendent is laready present it wont loop over , but we do need to check into deeper level 
https://github.com/Zimbra/zm-api-js-client/pull/447/files#diff-2f62190d7ad815da92e5dcf270fdd6e290f3362cf453d2951e2f4d7ed46e2badR182-R183
t

silentsakky (approved):

lavegupta (approved):

PR #448, opened 2021-02-25
PREAPPS-5636: Added cross-fetch module to support fetch api for nodejs project

sachinpatel88:


src/request/index.ts
```diff
@@ -13,7 +13,7 @@ import {
 	SingleBatchRequestResponse,
 	SOAPHeader
 } from './types';
-
+import fetch from 'cross-fetch';
```
silentsakky, line 16: how does this work in browsers? we don't want to include anything for browsers

src/request/index.ts
```diff
@@ -243,6 +243,10 @@ export function jsonRequest(
 		header.context.csrfToken = requestOptions.csrfToken;
 	}
 
+	if (requestOptions.authToken) {
+		options.headers['Cookie'] = `ZM_AUTH_TOKEN=${requestOptions.authToken}`;
```
silentsakky, line 247: this should not happen in case of browsers

src/request/index.ts
sachinpatel88: As per [cross-fetch documentation](https://www.npmjs.com/package/cross-fetch#install), including `cross-fetch` like this will [ponyfill](https://github.com/sindresorhus/ponyfill#ponyfill). So it won't modify original fetch api.

src/request/index.ts
sachinpatel88: Right. `requestOptions.authToken` will always be `undefined` except for offline-sync where we are passing explicitly this with value.

silentsakky (approved):

PR #452, opened 2021-03-16
PREAPPS-5867 : Adding HAB APIs.

BhavinBathani:
We are going to allow user to choose contacts from HAB with new contact-chooser dialog.
HAB related API calls needs to be added for ModernUI.
GetHABRequest API is implmented.

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -34,6 +34,8 @@ import {
 	GetDocumentShareURLResponseEntity,
 	GetFolderRequest as GetFolderRequestEntity,
 	GetRightsRequest,
+	HabGroup,
+	DlGroupMember,
 	InviteReply,
 	MessageInfo,
 	SaveDocument,
@@ -191,6 +193,19 @@ const hasUnreadDescendent = (folder: any): any => {
 
 	return false;
 };
+
+const updateGroupName = (habGroup: any) => ({
+	...habGroup,
+	name: get(habGroup, 'attributes.displayName')
+});
+
+const updateGroupNameRecur = (habGroups: any) => (
+	habGroups.map((habGroup: any) => {
+		habGroup = updateGroupName(normalize(HabGroup)(habGroup));
+		habGroup.habGroup && (habGroup.habGroups = [...updateGroupNameRecur(habGroup.habGroup)]);
+		return habGroup;
+	})
+);
 
 const setUnreadDescendentFlag = (folder: any) => {
 	const folderArray = get(folder, 'folders') || [];
@@ -968,6 +983,34 @@ export class ZimbraBatchClient {
 		}).then(res =>
 			normalize(Filter)(get(res, 'filterRules.0.filterRule') || [])
 		);
+
+	public getHAB = (habRootGroupId: string) =>
+		this.jsonRequest({
+			name: 'GetHAB',
+			body: {
+				habRootGroupId
+			},
+			namespace: Namespace.Account
+		}).then(res => {
+			const habGroups = get(res, 'ou.0');
+			return {
+				...habGroups,
+				habGroups : [...updateGroupNameRecur(habGroups.habGroup)]
+			};
+		});
+
+	public getDistributionListMembers = (limit: String, offset: String, dl: String) =>
+		this.jsonRequest({
+			name: 'GetDistributionListMembers',
+			body: {
+				dl : {
+					_content: dl
+				},
+				limit,
+				offset
+			},
+			namespace: Namespace.Account
+		}).then(res => normalize(DlGroupMember)(get(res, 'groupMembers.0.groupMember')) || []);
 
 	public getFolder = (options: GetFolderOptions) => {
 		return this.jsonRequest({
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -360,6 +360,14 @@ Folder.addMapping({
 });
 export { Folder };
 
+export const HabGroup = new Entity({
+	_attrs: 'attributes'
+});
+
+export const DlGroupMember = new Entity({
+	_attrs: 'attributes'
+});
+
 const ForwardMessageInput = new Entity({
 	e: ['emailAddresses', MailItemEmailAddress],
 	mp: ['mimeParts', MimePart],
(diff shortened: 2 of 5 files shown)


auminpatel (approved):

src/schema/schema.graphql
```diff
@@ -3102,6 +3149,14 @@ type Query {
 		traverseMountpoints: Boolean
 		folder: GetFolderFolderInput
 	): Folder
+	getHAB(
+		habRootGroupId: ID
+	): HabGroup
+	getDistributionListMembers(
+		limit: Int,
```
itsmeprashant, line 3156: Do you want it to change to Float? 
This question is in the context of the issue we faced recently with our customer.

itsmeprashant (approved):

src/schema/schema.graphql
silentsakky: I doubt we will pass limit which is very high

src/schema/schema.graphql
```diff
@@ -1125,6 +1125,43 @@ type Folder {
...
+
+type DlGroupMembers {
+	groupMember: [DlGroupMember]
+}
+
+type HabGroup {
+	name: String
+	id: String
+	parentHabGroupId: ID
+	seniorityIndex: Int
```
silentsakky, line 1147: can this contain value which is larger then 2lac, if yes then convert it to Float

src/batch-client/index.ts
```diff
@@ -967,6 +967,32 @@ export class ZimbraBatchClient {
...
 
+	public getHAB = (habRootGroupId: string) =>
+		this.jsonRequest({
+			name: 'GetHAB',
+			body: {
+				habRootGroupId
+			},
+			namespace: Namespace.Account
+		}).then(res => {
+			return get(res, 'ou.0') || [];
```
silentsakky, line 978: implicit return

src/schema/schema.graphql
```diff
@@ -1125,6 +1125,43 @@ type Folder {
...
 
+type DlAttrs {
+	commonName: String
+	displayName: String
+	lastName: String
+}
+
+type DlGroupMember {
+	name: String
+	_attrs: DlAttrs
```
silentsakky, line 1136: normalize to attributes

src/schema/schema.graphql
```diff
@@ -1125,6 +1125,43 @@ type Folder {
...
+type DlGroupMembers {
+	groupMember: [DlGroupMember]
+}
+
+type HabGroup {
+	name: String
+	id: String
+	parentHabGroupId: ID
+	seniorityIndex: Int
+	_attrs: HabGroupAttrs
```
silentsakky, line 1148: same as above

src/schema/schema.graphql
BhavinBathani: Limit would never be that high, so keeping it to Int only. I did use Int by following Zimbra API Docs.

src/schema/schema.graphql
BhavinBathani: seniorityIndex would never be that larger, so keeping it to Int only. I did use Int by following Zimbra API Docs.

src/batch-client/index.ts
BhavinBathani: Fixed.

src/schema/schema.graphql
BhavinBathani: Fixed.

src/schema/schema.graphql
BhavinBathani: Fixed.

madhavdhuppe (approved):

src/batch-client/index.ts
```diff
@@ -192,6 +194,21 @@ const hasUnreadDescendent = (folder: any): any => {
 	return false;
 };
 
+const updateGroupName = (habGroup: any) => {
+	return {
```
silentsakky, line 198: implicit return

src/batch-client/index.ts
```diff
@@ -192,6 +194,21 @@ const hasUnreadDescendent = (folder: any): any => {
...
+		...habGroup,
+		name: get(habGroup, 'attributes.displayName')
+	};
+}
+
+const updateGroupNameRecur = (habGroups: any) => {
+	return habGroups.map((habGroup: any) => {
```
silentsakky, line 205: implicit return

lavegupta (approved):

src/batch-client/index.ts
BhavinBathani: Fixed.

src/batch-client/index.ts
BhavinBathani: Fixed.

src/normalize/entities.ts
```diff
@@ -360,6 +360,48 @@ Folder.addMapping({
...
+
+HabGroup.addMapping({
+	_attrs: ['attributes', HabGroupAttrs],
+	habGroup: [ 'habGroups', HabGroup ]
+});
+
+export { HabGroup };
+
+export const DlAttrs = new Entity({
+	commonName: 'commonName',
```
silentsakky, line 391: entities which are same should not be put here

BhavinBathani: Fixed.

src/batch-client/index.ts
```diff
@@ -153,6 +155,7 @@ import {
 
 import { CASTING_PREFS } from './constants';
 import { Notifier } from './notifier';
+import { some } from 'lodash';
```
silentsakky, line 158:
```suggestion
import some from 'lodash/some';
```

silentsakky: when you use new method from lodash in api-client, you will have to modify rollup.config.js to consider it as external dependency

BhavinBathani: Removed as it is unused. may be added automatically.

src/batch-client/index.ts
```diff
@@ -969,6 +984,34 @@ export class ZimbraBatchClient {
...
 
+	public getHAB = (habRootGroupId: string) =>
+		this.jsonRequest({
+			name: 'GetHAB',
+			body: {
+				habRootGroupId
+			},
+			namespace: Namespace.Account
+		}).then(res => {
+			const habGroups = get(res, 'ou.0');
```
silentsakky, line 995: I don't see usage of normalize function here, don't we need it?

silentsakky (approved):

sachinpatel88 (approved):

PR #456, opened 2021-03-30
PREAPPS-5951 Add detailed logs for Sync process

damini-vashishtha:


damini-vashishtha:
closing this pull request, as it is not needed

PR #472, opened 2021-06-18
PREAPPS-6138: Handle attachments when ci / cd keys are not present in MIME

madhavdhuppe:
Set the contentDisposition in the attachment item as per their classification while normalizing mime parts.

### DIFF
--- a/src/utils/normalize-mime-parts.ts
+++ b/src/utils/normalize-mime-parts.ts
@@ -77,9 +77,21 @@ export function normalizeMimeParts(
 		isDesktop
 	}: { isDesktop?: string; jwtToken?: string; origin?: string }
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = (
+		{ ...attachment },
+		forcedContentDisposition?: string
+	) => {
 		attachment.messageId = attachment.messageId || message.id;
 		attachment.url = getAttachmentUrl(attachment, { origin, jwtToken });
+
+		if (!attachment.contentDisposition) {
+			if (forcedContentDisposition === 'attachments') {
+				attachment.contentDisposition = 'attachment';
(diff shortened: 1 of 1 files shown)


src/utils/normalize-mime-parts.ts
```diff
@@ -77,9 +77,19 @@ export function normalizeMimeParts(
...
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = (
+		{ ...attachment },
+		forcedContentDisposition?: any
+	) => {
 		attachment.messageId = attachment.messageId || message.id;
 		attachment.url = getAttachmentUrl(attachment, { origin, jwtToken });
+
+		if (forcedContentDisposition === 'attachments') {
```
silentsakky, line 87: this should be added only when original attachment data doesn't have content-disposition header
we shouldn't change value if already present

src/utils/normalize-mime-parts.ts
```diff
@@ -77,9 +77,19 @@ export function normalizeMimeParts(
 		isDesktop
 	}: { isDesktop?: string; jwtToken?: string; origin?: string }
 ) {
-	const processAttachment = ({ ...attachment }) => {
+	const processAttachment = (
+		{ ...attachment },
+		forcedContentDisposition?: any
```
silentsakky, line 82: type should be string instead of any

src/utils/normalize-mime-parts.ts
madhavdhuppe: Actually we are doing the same now as well but will add conditional code for betterment

src/utils/normalize-mime-parts.ts
madhavdhuppe: sure

silentsakky (approved):

BhavinBathani (approved):

PR #489, opened 2021-11-29
PREAPPS-6193 : Display signed/encrypted message properly.

BhavinBathani:
Receiver will be able to read signed/encrypted message.
Server does the decryption and provide decrypted content in case of server-smime.
Display info that the message was Signed/Encrypted while viewing email.
Display info if server responded with exception.

### DIFF
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -198,6 +198,43 @@ MimePart.addMapping({
 	attach: ['attachments', AttachmentsInfo]
 });
 
+const SmimeCertsSubjectRfc822Name = new Entity({
+	_content: 'content'
+});
+
+const SmimeCertsSubjectAltName = new Entity({
+	rfc822Name: ['rfc822Name', SmimeCertsSubjectRfc822Name]
+});
+
+const commonSmimeCertsFields = {
+	c: 'country',
+	cn: 'commonName',
+	o: 'organizationName',
+	st: 'state'
+};
+
+const SmimeCertsIssuedBy = new Entity({
+	...commonSmimeCertsFields,
+	l: 'locality'
+});
+
+const SmimeCertsIssuedTo = new Entity({
+	...commonSmimeCertsFields,
+	ou: 'organizationUnit'
+});
+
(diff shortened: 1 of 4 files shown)


src/normalize/entities.ts
```diff
@@ -198,6 +198,44 @@ MimePart.addMapping({
...
+});
+
+const SmimeCertsIssuedTo = new Entity({
+	c: 'country',
+	cn: 'commonName',
+	emailAddress: 'emailAddress',
+	o: 'organizationName',
+	ou: 'organizationUnit',
+	st: 'state'
+});
```
lavegupta, line 225: `
const commonSmimeCertsFields = {
	c: 'country',
	cn: 'commonName',
	emailAddress: 'emailAddress',
	o: 'organizationName',
	st: 'state'
};

const SmimeCertsIssuedBy = new Entity({
    ...commonSmimeCertsFields,
	l: 'locality'
});

const SmimeCertsIssuedTo = new Entity({
    ...commonSmimeCertsFields,
	ou: 'organizationUnit'
});`

src/normalize/entities.ts
```diff
@@ -198,6 +198,44 @@ MimePart.addMapping({
...
+	emailAddress: 'emailAddress',
+	l: 'locality',
+	o: 'organizationName',
+	st: 'state'
+});
+
+const SmimeCertsIssuedTo = new Entity({
+	c: 'country',
+	cn: 'commonName',
+	emailAddress: 'emailAddress',
```
silentsakky, line 221: no need to specify if key and value are same

silentsakky (approved):

src/normalize/entities.ts
BhavinBathani: Fixed.

src/normalize/entities.ts
BhavinBathani: Fixed.

silentsakky (approved):

k-kato (approved):

lavegupta (approved):

silentsakky (approved):

PR #496, opened 2021-12-27
ZBUG-2363 Modern UI, message body displaying partially

subrahmanyaShetty090:
-Added truncated property in mime parts

### DIFF
--- a/src/normalize/entities.ts
+++ b/src/normalize/entities.ts
@@ -13,9 +13,7 @@ const MimePart = new Entity({
 	cl: 'contentLocation',
 	ct: 'contentType',
 	s: 'size',
-	part: 'part',
-	mid: 'messageId',
-	content: 'content'
+	mid: 'messageId'
 });
 
 const CalendarItemAlarmTriggerRelative = new Entity({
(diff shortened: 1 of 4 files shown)


damini-vashishtha (approved):

src/schema/schema.graphql
```diff
@@ -1975,6 +1975,7 @@ type MimePart {
 	url: String
 	messageId: ID
 	base64: String
+	truncated:Boolean
```
silentsakky, line 1978: space after colon

subrahmanyaShetty090: implemented this change.

silentsakky (approved):

src/normalize/entities.ts
```diff
@@ -15,7 +15,8 @@ const MimePart = new Entity({
 	s: 'size',
 	part: 'part',
 	mid: 'messageId',
-	content: 'content'
+	content: 'content',
+	truncated: 'truncated'
```
silentsakky, line 19: we don't need to add these if key/value are same

silentsakky: also remove part key from line no 16

subrahmanyaShetty090: removed these key value pair.

silentsakky (approved):

PR #497, opened 2021-12-28
PREAPPS-1045: Added messages entity in the Attachment

miteshsavani810:


src/schema/schema.graphql
```diff
@@ -2016,6 +2016,7 @@ input DocumentInput {
 input AttachmentInput {
 	attachmentId: String
 	documents: [DocumentInput]
+	messages: [DocumentInput]
```
silentsakky, line 2019: should we rename DocumentInput in this case as we are using for both document and message

silentsakky: or create separate entity for MessageInput

miteshsavani810: Done.

BhavinBathani (approved):

silentsakky (approved):

damini-vashishtha (approved):

silentsakky (approved):

PR #503, opened 2022-02-10
ZBUG-1922 Shared folder Navigation issue in ModernUI

SohamBiswasTelus:
-Made changes in Zimbra-in-memory-cache.ts for creating unique id for caching.

-Made changes in batch-client/index.ts
In updateAbsoluteFolderPath

### DIFF
--- a/src/apollo/zimbra-in-memory-cache.ts
+++ b/src/apollo/zimbra-in-memory-cache.ts
@@ -22,10 +22,12 @@ const dataIdFromObject = (object: any) => {
 		case 'Folder':
 			if (object.local && object.local !== null) {
 				return `${defaultDataIdFromObject(object)}:local`;
-			}
-			if (object.id === '1') {
+			} else if (object.id === '1') {
 				// Cache the root folder based on both ID and UUID from server
 				return `${object.__typename}:${object.id}:${object.uuid}`;
+			} else if (object.id?.includes(':')) {
+				// Generating unique ID for Caching the mounting folders based on ID and absFolderPath
+				return `${object.id}:${object.absFolderPath}`;
 			}
 			return defaultDataIdFromObject(object);
 		case 'AutoCompleteMatch':
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -233,13 +233,12 @@ function updateAbsoluteFolderPath(originalName: any, parentFolderAbsPath: string
 		if (originalName === 'USER_ROOT') {
 			folder.absFolderPath = `${parentFolderAbsPath}${folder.absFolderPath}`;
 		} else {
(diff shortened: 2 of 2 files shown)


src/apollo/zimbra-in-memory-cache.ts
```diff
@@ -22,10 +22,13 @@ const dataIdFromObject = (object: any) => {
...
 				return `${defaultDataIdFromObject(object)}:local`;
-			}
-			if (object.id === '1') {
+			} else if (object.id === '1') {
 				// Cache the root folder based on both ID and UUID from server
 				return `${object.__typename}:${object.id}:${object.uuid}`;
+			} else if (object?.id?.includes(':')) {
+				// Cache the root folder based on both ID and UUID from server
+				let i = `${object.id}:${object.absFolderPath}`;
+				return i;
```
silentsakky, line 31: - [ ] combine both statements
- [ ] also fix comment
- [ ] absFolderPath can contain special characters, which can break caching mechanism, how would you handle that?

src/batch-client/index.ts
```diff
@@ -233,13 +233,12 @@ function updateAbsoluteFolderPath(originalName: any, parentFolderAbsPath: string
 		if (originalName === 'USER_ROOT') {
 			folder.absFolderPath = `${parentFolderAbsPath}${folder.absFolderPath}`;
 		} else {
-			folder.absFolderPath = folder.absFolderPath.replace(`/${originalName}`, parentFolderAbsPath);
+			folder.absFolderPath = `${parentFolderAbsPath}/${folder?.name}`;
```
silentsakky, line 236: do we need optional chaining operator here?

SohamBiswasTelus: I have changed the code.

src/apollo/zimbra-in-memory-cache.ts
SohamBiswasTelus: First two comments code change is done.
For third point I have checked with `${{(PAE??SI**~1!@#%<><>))+^==^` seems like working fine. And we cannot use colon, slash and double-quote in name

src/apollo/zimbra-in-memory-cache.ts
```diff
@@ -22,10 +22,12 @@ const dataIdFromObject = (object: any) => {
 		case 'Folder':
 			if (object.local && object.local !== null) {
 				return `${defaultDataIdFromObject(object)}:local`;
-			}
-			if (object.id === '1') {
+			} else if (object.id === '1') {
 				// Cache the root folder based on both ID and UUID from server
 				return `${object.__typename}:${object.id}:${object.uuid}`;
+			} else if (object?.id?.includes(':')) {
```
silentsakky, line 28: not sure if we need optional chaining operator here

silentsakky (approved):

PR #506, opened 2022-02-25
ZBUG-2332: Read recipient response as an attachment

miteshsavani810:
### Problem: 
Modern UI shows inessential attachment with Unnamed on Read-receipt response message, Folder Share message and Undelivered message response.

### Solution:
Ignoring the message part to be consider as attachment based on content type

### DIFF
--- a/src/utils/normalize-mime-parts.ts
+++ b/src/utils/normalize-mime-parts.ts
@@ -1,3 +1,12 @@
+/** List of content type not show as attachment */
+const ignoreContentTypeToShowAsAttachment = [
+	'application/pkcs7-signature',
+	'application/x-pkcs7-signature',
+	'message/delivery-status', // present in Undelivered mail
+	'message/disposition-notification', // present in read-receipt response
+	'xml/x-zimbra-share' // present in folder share message
+];
+
 function normalizeCid(cid: string) {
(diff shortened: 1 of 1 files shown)


src/utils/normalize-mime-parts.ts
```diff
@@ -1,3 +1,12 @@
+/** List of content type not show as attachment */
+const ignoreContentTypeToShowAsAttachment = [
+	'application/pkcs7-signature',
+	'application/x-pkcs7-signature',
+	'message/delivery-status', // present in Undelivered mail
+	'message/disposition-notification', // present in read-receipt response
+	'xml/x-zimbra-share' // present in folder share message
```
silentsakky, line 7: if I remember correctly, there is ZBUG for this case also, can you find that and link with this ticket?

silentsakky (approved):

src/utils/normalize-mime-parts.ts
silentsakky: https://jira.corp.synacor.com/browse/ZBUG-2452

silentsakky: please get ticket validated with this change

lavegupta (approved):

PR #507, opened 2022-03-04
ZBUG-1926: No option to select persona in the From field in modern UI

miteshsavani810:
- Requesting payload with `sendAsDistList` and `sendOnBehalfOfDistList` right on `DiscoverRights` request.

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -118,6 +118,7 @@ import {
 	ChangePasswordOptions,
 	CreateFolderOptions,
 	CreateSearchFolderOptions,
+	DiscoverRightOptions,
 	ExternalAccountDeleteInput,
 	ExternalAccountModifyInput,
 	FreeBusyOptions,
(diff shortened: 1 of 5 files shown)


src/batch-client/index.ts
```diff
@@ -764,6 +764,12 @@ export class ZimbraBatchClient {
 					},
 					{
 						_content: 'sendOnBehalfOf'
+					},
+					{
+						_content: 'sendAsDistList'
```
silentsakky, line 769: isn't this something we need to send through graphql instead of hardcoding here?

miteshsavani810: yes we do that. in that case we should send all this 4 rights from graphql.

miteshsavani810: Done.

damini-vashishtha (approved):

silentsakky (approved):

PR #508, opened 2022-03-09
ZBUG-2426: SAML SP-initiated logout does not work - zimbraWebClientLogoutURL

miteshsavani810:
- Created sendBeacon request to terminate session. This request will be called from zm-x-web directly when 'beforeunload' event trigger. See this https://github.com/ZimbraOS/zm-x-web/pull/3829/files#diff-e422e4a89b3245e4203985dfafed63e4e7a9abcb7bfc93c6ed161e75d36aeb9bR5

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -541,7 +541,7 @@ export class ZimbraBatchClient {
 			},
 			singleRequest: true,
 			namespace: Namespace.Account
-		}).then(res => normalize(ClientInfoResponse)(res));
+		}).then(res => normalize(ClientInfoResponse)(mapValuesDeep(res, coerceStringToBoolean)));
 
 	public contactAction = (options: ActionOptions) => this.action(ActionType.contact, options);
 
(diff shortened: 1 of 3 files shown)


src/batch-client/index.ts
```diff
@@ -854,6 +854,41 @@ export class ZimbraBatchClient {
 			singleRequest: true
 		});
 
+	public endSessionBecoanRequest = (options: JsonRequestOptions) => {
```
BhavinBathani, line 857: correct "beacon" spelling please

miteshsavani810: Done.

silentsakky (approved):

src/schema/schema.graphql
```diff
@@ -485,6 +485,7 @@ interface MailItem {
 type ClientInfoAttributes {
 	zimbraWebClientLoginURL: String
 	zimbraWebClientLogoutURL: String
+	zimbraWebClientSkipLogoff: String
```
silentsakky, line 488: should this be boolean?

silentsakky: that is true for all boolean types, still we convert it to boolean so we can use strict type checking of graphql

miteshsavani810: Done. Also updated in zm-x-web. see https://github.com/ZimbraOS/zm-x-web/pull/3829/commits/f6880a4eff6b16dc37f33448d1c2e3f00d646ab1

lavegupta (approved):

silentsakky (approved):

PR #510, opened 2022-03-14
PREAPPS-6625 : Fix failing desktop automation failures.

BhavinBathani:


src/batch-client/index.ts
```diff
@@ -590,7 +590,7 @@ export class ZimbraBatchClient {
 	public createContact = (data: CreateContactInput) =>
 		this.jsonRequest({
 			name: 'CreateContact',
-			body: createContactBody(data),
+			body: createContactBody(data, this.localStoreClient),
```
silentsakky, line 593: we should pass boolean from here, instead of object

BhavinBathani: Fixed.

silentsakky (approved):

lavegupta (approved):

PR #520, opened 2022-04-12
PREAPPS-6565: Message not displayed completely in modern UI

lavegupta:


src/utils/normalize-mime-parts.ts
```diff
@@ -146,6 +148,9 @@ export function normalizeMimeParts(
 					}
 
 					isBody = true;
+				} else if (!bodyType && disposition === 'inline' && MORE_TEXT_TYPES.includes(type)) {
```
silentsakky, line 151: too many if/else conditions can we simplify a bit, I see inline disposition condition at multiple places

lavegupta: Refactored. Please review.

miteshsavani810 (approved):

silentsakky (approved):

damini-vashishtha (approved):

PR #522, opened 2022-04-25
PREAPPS-6698:Modern UI - Users should not be allowed to use username in the password

sakshirawat0247:
-Added zimbraPasswordAllowUsername to the schema of AccountInfoAttrs

### DIFF
--- a/src/schema/generated-schema-types.ts
+++ b/src/schema/generated-schema-types.ts
@@ -128,6 +128,7 @@ export type AccountInfoAttrs = {
   zimbraMailBlacklistMaxNumEntries?: Maybe<Scalars['Int']>;
   zimbraMailQuota?: Maybe<Scalars['String']>;
   zimbraMtaMaxMessageSize?: Maybe<Scalars['Float']>;
+  zimbraPasswordAllowUsername?: Maybe<Scalars['Boolean']>;
   zimbraPasswordAllowedChars?: Maybe<Scalars['String']>;
   zimbraPasswordAllowedPunctuationChars?: Maybe<Scalars['String']>;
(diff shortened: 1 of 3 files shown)


src/schema/generated-schema-types.ts
```diff
@@ -129,6 +129,7 @@ export type AccountInfoAttrs = {
   zimbraMtaMaxMessageSize?: Maybe<Scalars['Float']>;
   zimbraPasswordAllowedChars?: Maybe<Scalars['String']>;
   zimbraPasswordAllowedPunctuationChars?: Maybe<Scalars['String']>;
+	zimbraPasswordAllowUsername?: Maybe<Scalars['Boolean']>;
```
silentsakky, line 132: is this manually added? this shouldn't have indentation issue as this is auto generated file

sakshirawat0247: updated the file again with correct auto generation

silentsakky (approved):

src/schema/schema.graphql
```diff
@@ -1310,6 +1310,7 @@ type AccountInfoAttrs {
 	zimbraFeatureMobileSyncEnabled: Boolean
 	zimbraFeatureRelatedContactsEnabled: Boolean
 	zimbraFeatureMailForwardingInFiltersEnabled: Boolean
+	zimbraPasswordAllowUsername: String
```
silentsakky, line 1313: should be boolean

sakshirawat0247: Yup, Changed it.Also figured out that we have a function coerceStringToBoolean which wasn't being used for resetPassword Response.So, added the same.

lavegupta (approved):

silentsakky (approved):

PR #528, opened 2022-05-27
PREAPPS-5640: Versioning files in Briefcase

miteshsavani810:


src/batch-client/index.ts
```diff
@@ -1352,6 +1353,29 @@ export class ZimbraBatchClient {
...
+				doc: {
+					id,
+					ver: version,
+					count
+				}
+			}
+		}).then(response => {
+			const data = normalize(ListDocumentRevisions)(response);
+			const versionIds = data.documents.map((doc: any) => doc.version);
+			const maxVersion = Math.max(...versionIds);
```
silentsakky, line 1370: I think version ids are in string format, how does Math.max works with string data? does it convert to int implicitly?
for clarity we should convert these version ids to int ourself

src/batch-client/index.ts
```diff
@@ -1352,6 +1353,29 @@ export class ZimbraBatchClient {
...
+			const latestDocumentVersion = data.documents.find((doc: any) => doc.version === maxVersion);
```
silentsakky, line 1371: is maxVersion a string or int?

src/batch-client/index.ts
```diff
@@ -1352,6 +1353,29 @@ export class ZimbraBatchClient {
...
+
+			return {
+				...latestDocumentVersion,
+				subchilds: data.documents
```
silentsakky, line 1375: instead of subchilds call it docs

src/schema/generated-schema-types.ts
```diff
@@ -3530,6 +3537,13 @@ export type QueryGetWorkingHoursArgs = {
 };
 
 
+export type QueryListDocumentRevisionsArgs = {
+  count: Scalars['String'];
```
silentsakky, line 3541: should be a number?

src/schema/generated-schema-types.ts
```diff
@@ -3530,6 +3537,13 @@ export type QueryGetWorkingHoursArgs = {
...
+  id: Scalars['ID'];
+  version: Scalars['String'];
```
silentsakky, line 3543: should be a number?

src/batch-client/index.ts
miteshsavani810: Version ids are in Int format.

src/batch-client/index.ts
miteshsavani810: it's int.

src/schema/generated-schema-types.ts
miteshsavani810: Done.

src/schema/generated-schema-types.ts
miteshsavani810: Done.

silentsakky (approved):

PR #530, opened 2022-06-02
PREAPPS-5118 Warn user for reaching zimbraMailSignatureMaxLength while saving signatures

damini-vashishtha:


src/schema/schema.graphql
```diff
@@ -1354,6 +1354,7 @@ type AccountInfoAttrs {
 	zimbraMailAttachmentMaxSize: Float
 	zimbraMtaMaxMessageSize: Float
 	zimbraMailAlias: [String]
+	zimbraMailSignatureMaxLength: Int
```
silentsakky, line 1357: we should use Float datatype for this as Int has very less limit
https://github.com/Zimbra/zm-api-js-client/commit/6d429c1bc7588f6b158e8e1a59f4a7646ba3030e

miteshsavani810 (approved):

lavegupta (approved):

PR #531, opened 2022-06-06
PREAPPS-6786: Handle download/delete action for version file.

miteshsavani810:


lavegupta (approved):

src/schema/schema.graphql
```diff
@@ -2453,6 +2454,12 @@ input InviteReplyInput {
 	exceptId: InstanceDate
 }
 
+input PurgetRevisionInput {
+	id: ID! 
+	ver: Int!
+	includeOlderRevisions: Int
```
silentsakky, line 2460: shouldn't this be BOOLEAN?

miteshsavani810: SOAP api is accepting this flag value as '0|1' hence sending it as Int.
See this https://files.zimbra.com/docs/soap_api/9.0.0/api-reference/index.html

silentsakky (approved):

silentsakky (approved):

PR #558, opened 2023-02-10
PREAPPS-6726: Implement Save-Search for Contacts, Calendar events and Briefcase files

namita-uttarkar-telus:
Implement Save-Search for Contacts, Calendar events and Briefcase files

- Added types field for folders, which will be used to seggregate search folders

### DIFF
--- a/src/schema/schema.graphql
+++ b/src/schema/schema.graphql
@@ -1140,6 +1140,7 @@ type Folder {
 	broken: Boolean #shared folder link is broken or not
 	deletable: Boolean
 	unreadDescendent: Boolean
+	types: String
 }
 
 type DlAttrs {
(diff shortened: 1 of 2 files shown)


miteshsavani810 (approved):

src/schema/schema.graphql
```diff
@@ -1140,6 +1140,7 @@ type Folder {
 	broken: Boolean #shared folder link is broken or not
 	deletable: Boolean
 	unreadDescendent: Boolean
+	types: String
```
silentsakky, line 1143: this should be converted to an enum with valid values
check https://files.zimbra.com/docs/soap_api/9.0.0/api-reference/zimbraMail/GetFolder.html

namita-uttarkar-telus: for mail vertical type can go as "". and enum doesn't accept empty string. Getting error like this. index.js:207 Error: Enum type FolderView cannot include value: "".
Tried with null value as well. Getting same error.

silentsakky (approved):

PR #586, opened 2023-06-19
PREAPPS-7334: Edit the custom pills in the required format while creating any event

sakshirawat0247:
- added new value displayName to attendees details


sakshirawat0247:
As the implementation approach has changed, this Branch and piece of code is not required anymore, so closing it.

PR #591, opened 2023-07-05
PREAPPS-7359:  Disable Mail recall option if mail recall time is elapsed

namita-uttarkar-telus:
- Added zimraFeatureMailRecallEnabled and zimbraFeatureMailRecallTime LDAP attributes

### DIFF
--- a/src/schema/schema.graphql
+++ b/src/schema/schema.graphql
@@ -1400,6 +1400,8 @@ type AccountInfoAttrs {
 	zimbraFeatureImportFolderEnabled: Boolean
 	zimbraFeatureExportFolderEnabled: Boolean
 	zimbraFeatureGroupCalendarEnabled: Boolean
+	zimbraFeatureMailRecallEnabled: Boolean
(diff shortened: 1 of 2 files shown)


src/schema/schema.graphql
```diff
@@ -1400,6 +1400,8 @@ type AccountInfoAttrs {
 	zimbraFeatureImportFolderEnabled: Boolean
 	zimbraFeatureExportFolderEnabled: Boolean
 	zimbraFeatureGroupCalendarEnabled: Boolean
+	zimbraFeatureMailRecallEnabled: Boolean
+	zimbraFeatureMailRecallTime: Int
```
silentsakky, line 1404: what would be maximum value which soap api can give for this property? wanted to make sure we are not running into issues with data which Integer can't hold

namita-uttarkar-telus: Maximum value for zimbraFeatureMailRecallTime is 30 min

silentsakky (approved):

PR #594, opened 2023-07-24
PREAPPS-7390: Remove "/@zimbra" dependency from dev proxy

miteshsavani810:


src/batch-client/index.ts
```diff
@@ -1,4 +1,5 @@
 import DataLoader from 'dataloader';
+import { Agent } from 'https';
```
silentsakky, line 2: this is nodejs specific package, not sure we should use this in web code

miteshsavani810: Removed

lavegupta (approved):

damini-vashishtha (approved):

PR #613, opened 2023-11-03
PREAPPS-7532: Distribution list is missing on choose contact popup

sandesh-telus:
- Added new query to fetched distribution lists details

### DIFF
--- a/src/batch-client/index.ts
+++ b/src/batch-client/index.ts
@@ -934,6 +934,18 @@ export class ZimbraBatchClient {
 			singleRequest: true
 		});
 
+	public getAccountDistributionLists = (attrs: String, ownerOf: number) =>
+		this.jsonRequest({
+			name: 'GetAccountDistributionLists',
+			body: {
+				attrs,
+				ownerOf
+			},
+			namespace: Namespace.Account
+		}).then(res => {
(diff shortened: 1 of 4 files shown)


src/batch-client/index.ts
```diff
@@ -934,6 +934,18 @@ export class ZimbraBatchClient {
...
+		this.jsonRequest({
+			name: 'GetAccountDistributionLists',
+			body: {
+				attrs,
+				ownerOf
+			},
+			namespace: Namespace.Account
+		}).then(res => {
+			return { dls: res?.dl ? res.dl : [] };
+		});
```
lavegupta, line 947: `}).then(res => ({ dls: res?.dl || [] }));`

sandesh-telus: Done

miteshsavani810: Make it inline statement. as per suggestion it is still missing.

sandesh-telus: @miteshsavani810 sorry I missed that point. I have make it inline statement. Please verify.

miteshsavani810 (approved):

lavegupta (approved):

PR #639, opened 2023-12-21
PREAPPS-7505 added apollo-client as external dependancy

sitaram-mulik-synacor:


rollup.config.js
```diff
@@ -4,23 +4,21 @@ import graphql from '@rollup/plugin-graphql';
 import localResolve from 'rollup-plugin-local-resolve';
 import nodeResolve from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
-import pkg from './package.json';
+import { visualizer } from 'rollup-plugin-visualizer';
 
-let FORMAT = process.env.FORMAT;
+const FORMAT = process.env.FORMAT;
 
-// graphql-tools currently has a rollup build failure, so always call it an external until they fix it
```
sitaram-mulik-synacor, line 11: I have added back `graphql-tools` as default dependancy since we are not using it in zm-x-web. So it will be bundled again with the api-js-client. I did not see any failure with the build. @silentsakky let me know if this change is fine.

rollup.config.js
```diff
@@ -4,23 +4,21 @@ import graphql from '@rollup/plugin-graphql';
...
-// otherwise, make all npm production dependencies external, plus their subpath usages
-// throughout the codebase, which rollup doesn't automatically pick up on
-let external = FORMAT==='es' ?
```
sitaram-mulik-synacor, line 14: Not sure why we were added these external dependancies only for ESM modules. I have removed the condition, so the following dependancies will be added as external for commonJs and UMD modules as well. @silentsakky thoughts ?

rollup.config.js
```diff
@@ -34,9 +32,10 @@ export default {
 			extensions: ['.ts'],
 			exclude: 'node_modules/**',
 			babelHelpers: 'bundled'
-		})
+		}),
+		visualizer()
```
silentsakky, line 36: will this always create visualisation as part of the build? Ideally, we should do this only when a developer wants to do that.

rollup.config.js
silentsakky: I am not very sure why it was done that way. we might need to understand how these package systems work

rollup.config.js
silentsakky: change looks good but I have a question, how are we notifying the package user that there are certain runtime dependencies which needs to be satisfied to make it work. in normal scenarios we use peerDependencies for that but I am not sure we can do that here.

rollup.config.js
sitaram-mulik-synacor: With current setting, it will create a file with name stats.html after the build. The file will not open automatically. Will check if we have any option to disable it or make it explicit.

rollup.config.js
sitaram-mulik-synacor: Moving the dependancies form dependancy array to peerDependancies will result in build failure in our case, since it is not supported below npm version 7. So in that case we must add them in the devDependancies or keep then in the dependancies as it is. I though it will be confusing to keep the same packages at two different locations to serve same purpose.

rollup.config.js
sitaram-mulik-synacor: Will do some research on this.

sitaram-mulik-synacor: Tried using microbundle however will need more time to fix the issues like bundling the graphql files. Hence I am reverting this change for now so that it will behave like old build. Means the external dependancies will be added for only es modules.

rollup.config.js
sitaram-mulik-synacor: Fixed this. Added a separate command to generate build analysis data.

silentsakky (approved):

miteshsavani810 (approved):

PR #662, opened 2024-02-08
PREAPPS-7648: As the admin of a distribution list add ability to edit the Distribution List properties in Modern UI

sandesh-telus:
- Added mutation 'distributionListAction'. It will save the DL attributes data.


src/batch-client/index.ts
```diff
@@ -783,6 +784,13 @@ export class ZimbraBatchClient {
 			singleRequest: true
 		}).then(Boolean);
 
+	public distributionListAction = (dlActions: DistributionListActionInput) =>
+		this.jsonRequest({
+			name: 'DistributionListAction',
+			body: dlActions,
+			namespace: Namespace.Account
+		}).then(Boolean);
```
silentsakky, line 792: why do we need `.then(Boolean)` ?

src/schema/schema.graphql
```diff
@@ -3169,6 +3169,39 @@ input SaveDocumentInput {
 	document: SaveDocumentInput
 }
 
+input DLActionAttrInput {
+	n: String
```
silentsakky, line 3173: we usually try to use more readable keys in schema and normalize/denormalize it as part of data layer

src/batch-client/index.ts
sandesh-telus: I am getting below response from 'DistributionListAction' api. Hence passing boolean 
image

src/schema/schema.graphql
sandesh-telus: Done.

src/batch-client/index.ts
HimanshuSojitraSynacor: so how does passing `Boolean` help in this case? If we don't care about the response then we should not attach the `.then()` handler.

sandesh-telus: @HimanshuSojitraSynacor, If I removed the `.then(Boolen)` and set the return type of api to `String` like `distributionListAction(dlActions: DistributionListActionInput!): String` then I get below error.
image

HimanshuSojitraSynacor (approved):

damini-vashishtha (approved):

PR #670, opened 2024-02-21
PREAPPS-7671: Fetch and attach previous message body as per Composer mode format and zimbraPrefForwardReplyInOriginalFormat

miteshsavani810:
- Storing HTML message and Text message separately in the cache.

### DIFF
--- a/src/apollo/zimbra-in-memory-cache.ts
+++ b/src/apollo/zimbra-in-memory-cache.ts
@@ -40,7 +40,7 @@ const dataIdFromObject = (object: any) => {
 			if (object.part && object.part !== null) {
 				return `${defaultDataIdFromObject(object)}:${object.part}`;
 			}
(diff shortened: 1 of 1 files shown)


src/apollo/zimbra-in-memory-cache.ts
```diff
@@ -40,7 +40,7 @@ const dataIdFromObject = (object: any) => {
 			if (object.part && object.part !== null) {
 				return `${defaultDataIdFromObject(object)}:${object.part}`;
 			}
-			return defaultDataIdFromObject(object);
+			return `${defaultDataIdFromObject(object)}`;
```
silentsakky, line 43: do we need this?

miteshsavani810: No.
Previous changes are not require and there is nothing change hence Dropping the PR.

PR #744, opened 2025-01-20
PREAPPS-8183: Send and Archive option to support Inbox Zero

miteshsavani810:


src/schema/schema.graphql
```diff
@@ -2153,6 +2154,7 @@ input MailboxMetadataSectionAttrsInput {
 	zimbraPrefUndoSendEnabled: Boolean
 	zimbraPrefUndoSendTimeout: Int
 	archivedFolder: String
+	showSendAndArchive: String
```
silentsakky, line 2157: Boolean?

sitaram-mulik-synacor (approved):

src/schema/schema.graphql
miteshsavani810: Classic UI zimlet (com_zimbra_mailarchive) work with String. See classic ui code https://github.com/Zimbra/zm-zimlets/blob/275a90c84192bdaf69801c269a61c627dbd58fcb/src/zimlet/com_zimbra_mailarchive/mailarchive.js#L297
So when we have to disable "showSendAndArchive" feature then we need to store "showSendAndArchive" value as "false" (in lowercase).

lavegupta (approved):

damini-vashishtha (approved):

silentsakky (approved):

PR #759, opened 2025-06-10
ZBUG-4718: Fixed calendar checkbox action UI glitch issue

sitaram-mulik-synacor:
Issue: actual response was not being returned from checkCalendar function.

Solution:  Updated checkCalendar function to return to return actual result instead of static response.


src/batch-client/index.ts
```diff
@@ -383,7 +384,11 @@ export class ZimbraBatchClient {
 			}
 		});
 
-	public action = (type: ActionType, options: ActionOptions) => {
+	public action = async (
+		type: ActionType,
+		options: ActionOptions,
+		returnResponse: boolean = false
```
silentsakky, line 390: not sure, if we need this parameter, we can by default return the response

sitaram-mulik-synacor: Most of the graphql mutation actions in zm-x-web is expecting boolean in the response. Hence we cant change its return type directly. In fact seems like dynamically returning the response also breaking the other mutation apis. Will check other way for this.

src/batch-client/index.ts
```diff
@@ -395,7 +400,7 @@ export class ZimbraBatchClient {
 				}
 			},
 			singleRequest: true
-		}).then(Boolean);
+		}).then(response => (returnResponse ? response : Boolean));
```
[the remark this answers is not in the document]
sitaram-mulik-synacor: As explained in [this](https://github.com/Zimbra/zm-api-js-client/pull/759#discussion_r2137075902) comment, other mutation will break if we directly return response from this.action function.

src/batch-client/index.ts
sitaram-mulik-synacor: Moved this logic inside `checkCalendar` function only since it creates issues in other mutations.

damini-vashishtha (approved):

PR #816, opened 2026-05-26
PREAPPS-9181: Use separate attribute to save group by date preference in modern UI

sitaram-mulik-synacor:
Use separate attribute to save group by date preference in modern UI to avoid any conflict with classic UI and also keep it enabled by default.

### DIFF
--- a/src/schema/schema.graphql
+++ b/src/schema/schema.graphql
@@ -2141,7 +2141,7 @@ type MailboxMetadataAttrs {
 	zimbraPrefFoldersExpanded: String
 	zimbraPrefFolderTreeSash: Int
 	zimbraPrefGenerateLinkPreviews: Boolean
-	zimbraPrefGroupByList: String
+	zimbraPrefGroupByListModern: String
 	zimbraPrefMessageListDensity: String
 	zimbraPrefMultitasking: String
 	zimbraPrefReadingPaneSashHorizontal: Int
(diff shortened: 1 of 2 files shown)


lavegupta (approved):

k-kato (approved):

src/schema/schema.graphql
```diff
@@ -2141,7 +2141,7 @@ type MailboxMetadataAttrs {
 	zimbraPrefFoldersExpanded: String
 	zimbraPrefFolderTreeSash: Int
 	zimbraPrefGenerateLinkPreviews: Boolean
-	zimbraPrefGroupByList: String
+	zimbraPrefGroupByListModern: String
```
silentsakky, line 2144: does this mean, we will just ignore existing stored values in zimbraPrefGroupByList for users? that might become regression

sitaram-mulik-synacor: Technically we will ignore the stored values but I am not sure whether it will be treated as regression, because impact wise I see it as following:

Case 1: For users who enabled group by date using `zimbraPrefGroupByList` (storing value as "2:GROUPBY_DATE") -  they will not be impacted as it will be our default option. 

Case 2: For users who did not enable group by date using `zimbraPrefGroupByList` (No valid value stored) - they will start seeing mails grouped by date. However we are making this as default behavior, so not sure whether we can call it as regression or a expected change as per requirement.

Case 3: Users who first enabled and then disabled group by date using `zimbraPrefGroupByList` (storing value as "2:GROUPBY_NONE") - They will start seeing mails grouped by date. But again similar to case 2, not sure whether we can treat this as regression or an expected change. 

Case 4: Classic UI users - will not be impacted..

I cant think of any other scenario where it can be seen as regression let me know if I am missing anything.

silentsakky: ok understood, it's intentional change so we can document in release notes accordingly.

silentsakky (approved):

src/schema/schema.graphql
sitaram-mulik-synacor: Added comment in the ticket - https://synacor.atlassian.net/browse/PREAPPS-9181?focusedCommentId=402822

